PR #11, opened 2016-12-20
Checking in windows desktop automated test wrapper and example projects

PandaMagnus:


IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/SimpleWpfApp.AutomatedUiTests/Windows/SimpleWpfAppWindow.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        public bool CheckAllListItems(params string[] expectedItems)
+        {
+            if ( expectedItems == null )
+                throw new ArgumentNullException( nameof( expectedItems ) );
+
+            var existingListItems = ListItemsList;
+
+            if ( expectedItems.Length != ListItemsList.Count() )
```
Keboo: Why are you going back to the property when you have already pulled the value into a local variable `existingListItems`?

Keboo: There is an additional code smell here because it looks like you want to iterate over the enumerable twice. You probably want to do something like
```C#
List<WpfListItem> existingListItems = ListItemsList?.ToList();
if (expectedItems.Length != existingListItems?.Count)
```

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/SimpleWpfApp.AutomatedUiTests/Windows/SimpleWpfAppWindow.cs
```diff
@@ -0,0 +1,86 @@
...
+                throw new Exception("CheckAllListItems parameter count did not match the number of ListItems returned");
+            }
+            bool allExpectedItemsPresent = true;
+            int i = 0;
+
+            foreach ( var item in existingListItems )
+            {
+                try
+                {
+                    var test1 = item.GetChildren().FirstOrDefault().Name;
```
Keboo: In-used variable `test1`. It also can throw a null ref.

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/SimpleWpfApp/MainWindow.xaml
```diff
@@ -0,0 +1,34 @@
...
+        </Grid.RowDefinitions>
+        <StackPanel Grid.Row="0">
+            <TextBox Text="{Binding Path=MyListEntry}" AutomationProperties.AutomationId="autoTextInput" Margin="10,0"></TextBox>
+            <Button Content="Save" Command="{Binding Path=SaveCommand}" VerticalAlignment="Center" AutomationProperties.AutomationId="autoButtonSave" HorizontalAlignment="Center" Margin="0,5,0,0"/>
+        </StackPanel>
+        <Expander Grid.Row="1" AutomationProperties.AutomationId="autoExpanderList">
+            <ListBox ItemsSource="{Binding Path=MyList}" AutomationProperties.AutomationId="autoListBoxMyList">
+                <ListBox.ItemTemplate>
+                    <DataTemplate>
+                        <TextBlock>
```
Keboo: When you only have one pieces of text you can simplify this to be:
```xaml
<TextBlock Text="{Binding Path=MyListItem}" />
```

IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/DesktopControlHelpers.cs
```diff
@@ -0,0 +1,72 @@
...
+                    {
+                        foundControl = FindControlByName(control, c => new UITestControl(c), foundControl);
+                    }
+                }
+            }
+
+            return foundControl;
+        }
+    }
+}
```
Keboo: There is a bunch of code duplication between these two cases. I think you want something like this:
```C#
protected UITestControl BuildControlHeirarchy(SearchTypes searchType, string baseControl, params string[] controlList)        
{ 
    if (searchType == null) 
        throw new ArgumentNullException(nameof(searchType)); 
    if (baseControl == null) 
        throw new ArgumentNullException(nameof(baseControl)); 
    UITestControl foundControl; 
    switch(searchType)
    {
        case SearchTypes.AutomationId:
            foundControl = FindControlImpl((controlName, rootControl) => FindWpfControlByAutomationId(controlName, c => new WpfControl(c), rootControl)); 
        case SearchTypes.ControlName:
            foundControl = FindControlImpl((controlName, rootControl) => FindControlByName(controlName, c => new UITestControl(c), rootControl));
        default:
            foundControl = new UITestControl();
    }
    return foundControl; 
}

private UITestControl FindControlImpl(Func<string, UITestControl, UITestControl> findMethod, string baseControl, string[] childControls)
{
    UITestControl foundControl = findMethod(baseControl, FindWinWindowUnderTest()); 
    if ( childControls?.Any() == true) 
    { 
        foreach (string control in controlList) 
        { 
            foundControl = findMethod(control, foundControl);
        } 
    } 
    return foundControl;
}
```

Few things to note:
- `foundControl` starts as `null` and we only give it the default value in the unlikely case where we have not set it to anything else
- Use a switch rather than a bunch of if statements when acting on an enum.
- The common logic is abstracted into its own method
- Added null propogation operator when checking childControls. Typically you don't need to check values that come from params as the compiler will put in an empty array unless the caller specifically passes null.
- Zero testing of this code on my part, so there may be errors.

IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/GenericPlaybackSettings.cs
```diff
@@ -0,0 +1,34 @@
...
+            Playback.PlaybackSettings.MaximumRetryCount = 5;
+            Playback.PlaybackSettings.ShouldSearchFailFast = true;
+            Playback.PlaybackSettings.DelayBetweenActions = 100;
+            Playback.PlaybackSettings.SearchTimeout = 1000;
+            Playback.PlaybackSettings.MatchExactHierarchy = false;
+
+            //Tip from Stack Overflow to remove and then re-add error handling to get it to initialize properly
+            //Don't know why this works, but it did when Mike Curn checked using VS2010.
+            Playback.PlaybackError -= PlaybackErrorHandler;
+            Playback.PlaybackError += PlaybackErrorHandler;
```
Keboo: This is a common pattern to keep from accidently duplicating event handlers. I suspect that was likely the cause.

IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/Utilities.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        /// <summary>
+        /// Parse a text file and return a string of its contents
+        /// </summary>
+        /// <param name="path">The full path of the location and file name</param>
+        /// <returns>Returns the text of the chosen file</returns>
+        public string ParseTextFile(string path)
+        {
+            string text = null;
+            int seconds = 0;
```
Keboo: You probably want to name this something like `retryAttempts` since it is not actually a measurement of seconds.

IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/Utilities.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        /// Deletes a file given a location
+        /// </summary>
+        /// <param name="location">The location of the file without a slash at the end</param>
+        /// <param name="filename">The name and extension of the file</param>
+        public void DeleteDocument(string location, string filename)
+        {
+            try
+            {
+                File.Delete(location + "\\" + filename);
```
Keboo: Use Path.Combine here and you won't need to worry about if the `location` ends with a slash or not.

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        //Ordinarily this would contain more complex operations.
+        //Use this to abstract away some of the complexity to help the test author.
+        public void SaveDocument(string location, string filename)
+        {
+            //Check if doc with same name already exists in location before saving
+            string formattedPath = location + "\\" + filename;
+            if (File.Exists(formattedPath))
+            {
+                filename += "1";
```
PandaMagnus: It was not intentional. It was an oversight on me trying to come up with something quick and not double checking. However, your comment has got me thinking that there is a better way in general to handling saving files for a test. I will do some poking around with this...

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Tests/ExampleNotepadTest.cs
```diff
@@ -0,0 +1,44 @@
...
+        //These may be turned into data drive tests for conciseness and flexibility
+        [TestMethod]
+        [TestCategory("ExampleTest")]
+        public void NoCharactersEntered()
+        {
+            NotepadWindow.NotepadEdit.Text = Empty;
+            NotePadWindowHelpers.SaveDocument(SaveLocation, "testNoCharacters.txt");
+            var text = Utilities.ParseTextFile(SaveLocation + @"\testNoCharacters.txt");
+            Assert.AreEqual(Empty, text, "Unexpected text was found");
+            Utilities.DeleteDocument(SaveLocation, "testNoCharacters.txt");
```
PandaMagnus: I am undecided. I considered putting the DeleteDocument call in the test teardown, which will run even if there's a failure, but I wasn't sure if there would be value in being able to inspect the file should there be a failure. Same as above: I'll do some poking around with this to see if I can determine if one way is better than the other.

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/SimpleWpfApp.AutomatedUiTests/Windows/SimpleWpfAppWindow.cs
PandaMagnus: I think this answers my above question.

Keboo: See comment below. You likely want to simply make the local variable a list

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/SimpleWpfApp.AutomatedUiTests/Windows/SimpleWpfAppWindow.cs
PandaMagnus: Entirely removed. That was for some test I was doing to see how to get at certain data.

Keboo:
:shipit:

PR #12, opened 2017-02-02
Add desktop

PandaMagnus:
Changes based on review with Mark M.


IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/Utilities.cs
```diff
@@ -4,55 +4,49 @@
...
+                File.Delete(Path.Combine(location, fileName));
+                Thread.Sleep( 100 );
+                if (!File.Exists(Path.Combine(location, fileName)))
+                {
+                    break;
+                }
+                retryAttempts++;
+
+            } while ( retryAttempts < 10 );
+            if ( File.Exists(Path.Combine(location, fileName)) )
```
Keboo: Rather than doing the Path.Combine 3 times, just do it once at the beginning of the method and store the result in a local variable.
You can also update the XML docs, since it does not matter if the `location` parameter has the slash at the end.

IntelliTect.TestTools.WindowsTestWrapper/CustomDictionary.xml
```diff
@@ -0,0 +1,16 @@
...
+    <Recognized>
+
+      <Word>Intelli</Word>
+      <Word>Tect</Word>
+
+    </Recognized>
+
+  </Words>
+
+</Dictionary>
```
Keboo: It appears that there are multiple copies of this same dictionary. Rather than having one specific to each project, what about putting this file along side the solution file, and simply referencing it from both projects (ie. `<CodeAnalysisDictionary Include="..\CustomDictionary.xml" />`

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Tests/BaseTest.cs
```diff
@@ -10,25 +10,11 @@ namespace Notepad.AutomatedTests.Tests
     /// Summary description for BaseTest
     /// </summary>
     [CodedUITest]
-    public partial class BaseTest
+    public partial class BaseTest : BaseTestInherit
     {
+        public override string ApplicationLocation => Environment.SystemDirectory + "\\notepad.exe";
```
Keboo: Perhaps use `Path.Combine` here.

IntelliTect.TestTools.WindowsTestWrapper/CustomDictionary.xml
PandaMagnus: Hmmm... I thought I did that; there should only be a single file in the solution directory that gets referenced. I'll double check, though.

PandaMagnus:
@Keboo I ended up making some more changes to this PR based on some other comments from Mark. Want to give it another look before I close the PR?

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
```diff
@@ -10,6 +12,37 @@ public class NotepadWindowHelpers
...
+        public bool ValidateTextFile(string location, string fileName, string expectedText)
+        {
+            bool comparisonPass = false;
+            string combinedPath = Path.Combine( location, fileName );
+            if ( combinedPath == expectedText )
+            {
+                comparisonPass = true;
+            }
+            return comparisonPass;
+        }
```
Keboo: This method confuses me. 
The entire contents can be reduced to:
```C#
return Path.Combine(location, filename) == expectedText;
```
However if that is what is expected the name of the method and parameters should have their names updated since they seem to imply that it is checking the contents of the file.

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
```diff
@@ -10,6 +12,37 @@ public class NotepadWindowHelpers
...
+            {
+                File.Delete(combinedPath);
+                if (!File.Exists(combinedPath))
+                {
+                    break;
+                }
+                Thread.Sleep(100);
+                retryAttempts++;
+            } while (retryAttempts < 10);
+            if (File.Exists(combinedPath))
```
Keboo: You don't need the File.Exists checks here. If `File.Delete` does not throw an exception then you can assume that the file is not there.

IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/BaseTestInherit.cs
```diff
@@ -0,0 +1,58 @@
...
+            Playback.PlaybackSettings.ShouldSearchFailFast = true;
+            Playback.PlaybackSettings.DelayBetweenActions = 100;
+            Playback.PlaybackSettings.SearchTimeout = 1000;
+            Playback.PlaybackSettings.MatchExactHierarchy = false;
+
+            //Tip from Stack Overflow to remove and then re-add error handling to get it to initialize properly
+            //Don't know why this works, but it did when Mike Curn checked using VS2010.
+            Playback.PlaybackError -= PlaybackErrorHandler;
+            Playback.PlaybackError += PlaybackErrorHandler;
+            Debug.Write("Default test initialize finished");
```
Keboo: You probably want to use `Debug.WriteLine` here. Otherwise the Debug output for this method will put both messages on the same line.

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
PandaMagnus: Oh... duh. Changes made!

IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/BaseTestInherit.cs
PandaMagnus: Oh... duh. Changes made here, too!

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
PandaMagnus: I was thinking of cases where (and I've seen this before,) the test tries to delete the file before it's finished saving. Would you recommend a very short Sleep instead? I was originally catching exceptions and retrying, but Mark pointed out that that's quite dangerous as it originally made LOTS of assumptions that no exception would ever be valid. I'm trying to find a happy medium between not catching exceptions, but also not having a method that will fail just because of some sub-second timing oddness!

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
PandaMagnus: It would have helped to keep in the File.ReadAllText when I moved this over and refactored it a bit... That might have cleared up some confusion *facepalm*

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
Keboo: I mostly agree with @MarkMichaelis on the exceptions, however in this case I think it is by far preferable to this code. Because things outside of the program can make changes to the file system, the recommended approach is to attempt the IO action and check for any exceptions. The issue with the above code is if the file is still in the process of being saved, the code you have above, will throw an exception and you will never get to your retry logic (which is the exact case where you want to retry). 

From the [docs](https://msdn.microsoft.com/en-us/library/system.io.file.delete(v=vs.110).aspx) it would appears that there are two exception that could be thrown where you want to sleep and retry, IOException and UnauthorizedAccessException. The latter is the easy case because you can simply assume that if you seen that exception type you need to retry. The IOException is a little more tricky because to do it right you only want to retry in the cases where it is thrown because the other process still has a file handle open to it. For this we need to look a little closer at what [File.Delete](https://referencesource.microsoft.com/#mscorlib/system/io/file.cs,250) actually does. If you follow the code through you will end up [here](https://referencesource.microsoft.com/#mscorlib/system/io/__error.cs,135). Specifically notice that the error code that you care about is `Win32Native.ERROR_SHARING_VIOLATION`. This value is put into the HResult property of the IOException. So you also want to retry in the case where an IOException is thrown with HResult = 0x80070020 (look at the MakeHRFromErrorCode to see how I got this value). 
As a side note, this would be an excellent case to use the exception filters feature (keyword when in the catch) that was introduced in C# 6.0.

PandaMagnus: That is ridiculously helpful. I'll do some more updates based on the information and the code. Am I correct in assuming that if I do:
Try{ //action }
catch (UnauthorizedAccessException) { //retry stuff}
catch (Exception) {throw;}

Then the last catch statement will basically catch and throw everything else that isn't covered by the first one (or, first two, once I have one for catching the IOException)?

Keboo: Yes you are correct, but it is actually simpler than that. If you just want to re-throw the exception you can simply omit `catch(Exception){throw;}` and it will automatically throw anything that is not caught by your other catch blocks.

PandaMagnus: OKAY! There we go. Changes synced. I just learned way more about exceptions than I ever thought I would. :laughing:

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
```diff
@@ -15,16 +16,75 @@ public class NotepadWindowHelpers
...
+                        Debug.WriteLine($"Could not read file after {numberOfRetries} attempts.");
+                        throw;
+                    }
+                }
+                retryCount++;
+                Thread.Sleep(100);
+            } while (retryCount < numberOfRetries);
+
+
+            return result;
```
Keboo: Is it intentional that this would fall into the retry logic when the contents of the file do not match the expected text?

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
```diff
@@ -58,5 +119,10 @@ public static string AppendFileName(string filePath, string toAppend)
             var extension = Path.GetExtension(filePath);
             return Path.Combine(dir ?? "", string.Concat(fileName, toAppend, extension));
         }
+
+        private bool CheckForErrorSharingViolation(int hResult)
```
Keboo: Method can be made static

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
```diff
@@ -15,16 +16,75 @@ public class NotepadWindowHelpers
...
+            string combinedPath = Path.Combine(location, fileName);
+            int retryCount = 0;
+            int numberOfRetries = 10;
 
+            do
+            {
+                try
+                {
+                    File.Delete( combinedPath );
+                }
```
Keboo: You want a `return` or a `break` here otherwise it will just keep looping even in the success case.

IntelliTect.TestTools.WindowsTestWrapper/ExampleProjects/Notepad.AutomatedTests/Helpers/NotepadWindowHelpers.cs
PandaMagnus: Nope. I kinda sped through this and forgot to review. Thank you for the catch (along with the ones below!) I'll make the changes on lunch.

PR #13, opened 2017-02-09
Add desktop

PandaMagnus:
Moving solution file to main TestTools directory.
Suppressing code analysis warnings that should not be changed


Keboo:
:shipit:

PR #16, opened 2017-03-26
Upgraded the solution to refer to the csproj files

MarkMichaelis:
Switched sln to use csproj files rather than xproj files.


Keboo:
:shipit:

PR #17, opened 2017-08-07
Add find sibling

PandaMagnus:
Add method to find a control by its sibling in the control hierarchy
Adding test to utilize said method
Some general cleanup and a bug fix

### DIFF
--- a/IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/DesktopControls.cs
+++ b/IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/DesktopControls.cs
@@ -145,6 +145,17 @@ namespace IntelliTect.TestTools.WindowsTestWrapper
             return found;
         }
 
+        protected T FindControlBySibling<T>(Func<UITestControl, T> controlType, UITestControl siblingControl ) where T : UITestControl
+        {
+            if (controlType == null)
+                throw new ArgumentNullException(nameof(controlType));
+            if (siblingControl == null)
+                throw new ArgumentNullException(nameof(siblingControl));
+            T found = controlType(siblingControl);
+            found.SearchConfigurations.Add(SearchConfiguration.NextSibling);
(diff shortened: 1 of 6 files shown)


IntelliTect.TestTools.WindowsTestWrapper/IntelliTect.TestTools.WindowsTestWrapper/DesktopControls.cs
```diff
@@ -145,6 +145,17 @@ protected T FindWinControlByControlId<T>(string controlId, Func<UITestControl, T
...
         }
 
+        protected T FindControlBySibling<T>(Func<UITestControl, T> controlType, UITestControl siblingControl ) where T : UITestControl
+        {
+            if (controlType == null)
+                throw new ArgumentNullException(nameof(controlType));
+            if (siblingControl == null)
+                throw new ArgumentNullException(nameof(siblingControl));
+            T found = controlType(siblingControl);
+            found.SearchConfigurations.Add(SearchConfiguration.NextSibling);
```
Keboo: This code is making the assumption that the `controlType` delegate never returns null. Is that a valid assumption?

PandaMagnus: I think so. If the method is called with controlType or siblingControl being null, an exception is thrown. Beyond that, controlType is simply newing up a UiTestControl object (or derived class like WpfControl) with siblingControl as the parent.

WHICH... as I typed that, realized is a flaw in the logic. Sibling control should not be the parent. I will make some tweaks to fix this. :)

PandaMagnus: Oh wait, nevermind... I looked back through some old code and forgot that the NextSibling configuration is weird and totally does work this way (it wasn't just a fluke.)

Added a summary but otherwise I think assuming controlType never returns null is safe in this case.

PR #19, opened 2018-08-08
Initial commit with heaps of rough stuff

PandaMagnus:


IntelliTect.TestTools.Selenate/GoogleSearch/ExampleTests.cs
```diff
@@ -0,0 +1,57 @@
...
+            Assert.IsFalse(Browser.IsElementDisplayedAndExisting(Harness.SearchResultsDiv.By),
+                "Search results displayed when they were not expected");
+        }
+
+        [TestCleanup]
+        public void Teardown()
+        {
+            Browser.TakeScreenshot();
+            Browser.Driver.Quit();
+            Browser = null;
```
Keboo: Are these null sets needed? Won't the GC simply clean these up for you?

IntelliTect.TestTools.Selenate/GoogleSearch/GoogleBrowser.cs
```diff
@@ -0,0 +1,24 @@
+﻿using OpenQA.Selenium;
+using IntelliTect.TestTools.Selenate;
+
+namespace GoogleSearch
+{
+    class GoogleBrowser : Browser
```
Keboo: Minor thing: missing protection modifier.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+                        EnsureCleanSession = true,
+                        EnableNativeEvents = true,
+                        IgnoreZoomLevel = true,
+                        IntroduceInstabilityByIgnoringProtectedModeSettings = true,
+                        RequireWindowFocus = false
+                    };
+                    Driver = new InternetExplorerDriver(ieCaps);
+                    break;
+                default:
+                    throw new Exception($"Unknown browser: {browser}");
```
Keboo: Given that this exception is being thrown because of the passed in argument value. It would be better to throw an ArgumentException. Or possibly an ArgumentOutOfRangeException

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+            }
+
+            Driver.Manage().Window.Maximize();
+            Driver.Manage().Timeouts().PageLoad = TimeSpan.FromMinutes(2);
+        }
+
+        public void TakeScreenshot()
+        {
+            string fullPath = $"{Directory.GetCurrentDirectory()} + \\{((RemoteWebDriver)this.Driver).Capabilities.BrowserName}_{DateTime.Now}";
```
Keboo: When building up file paths. Favor using `Path.Combine` rather than string literals for slash.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+
+            Screenshot screenshot;
+
+            Console.WriteLine($"Saving screenshot to location: {fullPath}");
+            Directory.CreateDirectory(fullPath.Substring(0, fullPath.LastIndexOf("\\") + 1));
```
Keboo: Pretty sure what you want rather than the substring is `Path.GetDirectory`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// <summary>
+        ///Waits until a function evaluates to true OR times out after a specified period of time
+        /// </summary>
+        /// <param name="func">Function to evaluate</param>
+        /// <param name="secondsToWait">Secondes to wait until timeout / return false</param>
+        /// <returns></returns>
+        public bool WaitFor(Func<bool> func, int secondsToWait = 15)
+        {
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
```
Keboo: Stopwatch appears to be unused?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+            WebDriverWait wait = new WebDriverWait(Driver, TimeSpan.FromSeconds(secondsToWait));
+            try
+            {
+                wait.Until(SeleniumExtras.WaitHelpers.ExpectedConditions.ElementExists(by));
+            }
+            catch (WebDriverTimeoutException) { }
+            IReadOnlyCollection<IWebElement> elements = Driver.FindElements(by);
+            if (elements.Count > 0)
+            {
+                return elements.ElementAt(0).Displayed;
```
Keboo: Rather than `ElementAt(0)` you probably want `.First()`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+                    ex = e;
+                }
+                Task.Delay(500).Wait();
+            }
+            if (ex != null)
+                throw ex;
+            Task.Delay(500).Wait();
+        }
+
+        public Boolean SwitchWindow(string title)
```
Keboo: Use keyword `bool` in place of full class name `Boolean`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+        {
+            Exception ex = new Exception();
```
Keboo: Exception variable appears to be unused.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+                }
+
+                Task.Delay(500).Wait();
+            }
+            return false;
+        }
+
+        public IAlert Alert(int numberOfRetries = 50)
+        {
+            Exception ex = new Exception();
```
Keboo: `ex` variable appears un-used

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
```diff
@@ -0,0 +1,253 @@
...
+            return RetryAction("getproperty", propertyName);
+        }
+
+        public string GetCssValue(string propertyName)
+        {
+            return RetryAction("getcss", propertyName);
+        }
+
+        private const int DefaultRetryTime = 30;
+        private int TimeToRetry { get; set; } = DefaultRetryTime;
```
Keboo: I realize from usage that this implies seconds, but it might be better to either give it a more explicit name, or perhaps just use a `TimeSpan`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
```diff
@@ -0,0 +1,253 @@
...
+                catch (InvalidCastException e)
+                {
+                    ex = e;
+                }
+                catch (WebDriverException e)
+                {
+                    ex = e;
+                }
+
+                // Re-find the element to make sure it's valid.
```
Keboo: Why re-find the element if the method is just going to throw?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/WebElementExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+            element.FillInWith(value);
+            element.SendKeys(Keys.Tab);
+            // Wait afterward to let websites register the tab
+            Task.Delay(500).Wait();
+        }
+
+        public static void FillInWith(this IWebElement element, string value)
+        {
+            var count = 0;
+            while (element.GetAttribute("value") != value && count < 5)
```
Keboo: `count` is never incremented. Did you mean `while (element.GetAttribute("value") != value && count++ < 5)`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+            {
+                return elements.ElementAt(0).Displayed;
+            }
+            return false;
+        }
+
+        public void FrameSwitchAttempt(params By[] bys)
+        {
+            Exception ex = new Exception();
+            for (int i = 0; i < 50; i++)
```
Keboo: Is this just for doing retries? It appears to be duplicated in other methods. It would be good to extract the retry logic into its own method.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+            if (elements.Count > 0)
+            {
+                return elements.ElementAt(0).Displayed;
+            }
+            return false;
+        }
+
+        public void FrameSwitchAttempt(params By[] bys)
+        {
+            Exception ex = new Exception();
```
Keboo: Don't initialize a value you are not going to use. It would be better to simply do `Exception ex = null;`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,281 @@
...
+                    ex = e;
+                }
+                catch (NotFoundException e)
+                {
+                    ex = e;
+                }
+                Task.Delay(500).Wait();
+            }
+            if (ex != null)
+                throw ex;
```
Keboo: This is only going to throw the last exception. Is this expected/desired? If you want to throw all exceptions that occured, you could use an `AggregateException`

IntelliTect.TestTools.Selenate/GoogleSearch/ExampleTests.cs
PandaMagnus: Oops! Yes, it will clean those up. Was still in the mindset of the solution I pulled this out of, which was using DI to persist data across test runs. :)

IntelliTect.TestTools.Selenate/GoogleSearch/GoogleBrowser.cs
PandaMagnus: Fixed with next commit

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Fixed with next commit

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Fixed with next commit

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Fixed

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Fixed with next commit

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: This is one of the bigger "rough" things in here. I am going to refactor this method, FrameSwitchATtempt() and Alert() to conform more to the retrying for a specified period of time pattern instead of a number of attempts. I'll remove the reference (or use it) when I refactor the method.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Same as above.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
PandaMagnus: For cases where Selenium tried to pick up the element while it was loading. Basically, the use case of we were expecting the element to be there, but maybe it took a second longer to load than before, so we re-find it in the loop to give us a chance of picking it up after it's in the right state.

HOWEVER, when this loop was first written, we knew less about the element we were finding, so I'm thinking I might be able to offload the "re-find" to the start of the loop, and find it only if it's not in the correct state using a check on the Initialized property.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Yes and yes. Part of the effort to putting this into the IntelliTect repo is to take the several cases where people have added similar-ish retries and extract out the common logic. :) I'd prefer to defer that work for the next check-in as there's a lot to do, and I'd like to take it incrementally.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Fixed with next commit

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
Keboo: But I am not sure that is what this code does. The only way to get this far in the method is for an exception to be thrown and caught above (meaning the `ex` variable will be set). It then attempts this FindElement, which may or may not succeed and set the `WrappedElement` property. But regardless of whether this second find succeeds it will always hin the `if (ex != null)` below, and since we know that `ex` is set, this method will throw.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
Keboo: That's fine.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
PandaMagnus: I'm on my phone, so it's super hard to verify, but the throw *should* be outside the while loop to prevent that. If it's not, I will change it as soon as I get home and be extra diligent in checking my curly braces. 🙂

Keboo: So you are correct, the `throw` is outside of the `while` loop. However that doesn't actually fix the issue. So let's say you get here and do a second `FindElement` call and it doesn't throw an exception. That means we either exit the while loop or loop again.
If we exit the while loop, my above comment holds, and the method throws. If we loop again, the `ex` variable gets cleared, and we end up replacing the value we set in `WrappedElement`.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Oops, totally missed this... I'm just going to change it to a single screenshot folder. We did this initially to handle having multiple screenshots and video per test case, which I think for a something like this is overkill anyway.

Changed with next commit.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
PandaMagnus: Oooh, okay, makes sense. I'll continue ahead with looking into moving the re-find logic to the start of the loop and only if the element is in a bad state.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/WebElementExtensions.cs
PandaMagnus: Fixed with next commit

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Hmmm... I'm trying to figure out how to use AggregateException, but it looks like it can only be used when executing a task? Is that accurate?

Keboo: Something like this:
```C#
var exceptions = new List<Exception>();
...
exceptions.Add(ex);
...
throw new AggregateException(exceptions);
```

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,279 @@
...
+                    throw new ArgumentException($"Unknown browser: {browser}");
+            }
+
+            Driver.Manage().Window.Maximize();
+            Driver.Manage().Timeouts().PageLoad = TimeSpan.FromMinutes(2);
+        }
+
+        public void TakeScreenshot()
+        {
+            string fullPath = Path.Combine(Directory.GetCurrentDirectory(), "screenshot", ((RemoteWebDriver)this.Driver).Capabilities.BrowserName, "_", DateTime.Now.ToString());
```
Keboo: I think you may have gone a bit too far here. `Path.Combine` will put a slash between each of the parameters that you pass to it. To match what was there previously, you want to do normal string concatenation of the last several parameters.

PandaMagnus: Oops. 😬  I absolutely got overzealous in my usage here. Fixed with next commit.

PandaMagnus:
@Keboo  ALSO: Are you running analysis tools? I turned warnings to errors and had zero failures. But IIRC, analysis tools are a separate package for .Net Standard and maybe .Net Core?

Keboo:
@PandaMagnus 
In regards to the threading. Using Task.Wait() vs async methods actually does nothing in regards to tests being run synchronously or in parallel. That is a setting that is completely controlled by the test runner (in this case Visual Studio Test Tools). Making the method async and awaiting the long running calls merely allows your program to respond effectively when outside events occur (which when you are working with external processes and operating system things, outside events occur often). 

As for the analysis tools, I was referring to just the built-in compiler ones (yes there are additional analyzers you can get via nuget, but I was just looking at compiler ones). If you turn them on via the UI (Project Properties > Build > Treat warnings as errors = All) Visual Studio will only enable them for your current configuration. If you look at the PropertyGroup that gets added to the csproj you can simply move the change up into default property group so it will apply to all configurations. And yes, you are correct, there are currently no warnings that are triggering errors, but good to have it enabled as it will force everyone contributing to this to follow good coding practices.

PandaMagnus:
Thanks, Kevin! I should get a chance to make the rest of the changes tonight or tomorrow.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,275 @@
...
+                    throw new ArgumentException($"Unknown browser: {browser}");
+            }
+
+            Driver.Manage().Window.Maximize();
+            Driver.Manage().Timeouts().PageLoad = TimeSpan.FromMinutes(2);
+        }
+
+        public void TakeScreenshot()
+        {
+            string fullPath = Path.Combine(Directory.GetCurrentDirectory(), "screenshot", $"{((RemoteWebDriver)this.Driver).Capabilities.BrowserName}_{DateTime.Now.ToString()}");
```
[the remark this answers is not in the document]
PandaMagnus: Oops, didn't finish the changes to this before I committed. I'll pick this back up tomorrow.

PandaMagnus:
@Keboo  I might have to swing by the office to run some ideas by you. I'm trying to resolve the async/Task.Delay concerns, and I'm having trouble reconciling returning a Task<T> type, and doing normal selenium things with the return (but only in certain circumstances.)

I guess my main thought here is, I could technically remove the Task.Delays entirely from these methods; the main reason they're there is to prevent pegging the CPU with these retries on an unrestricted timeline, so is there a different/better way?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,251 @@
...
+                catch(AggregateException)
+                {
+                    return false;
+                }
+
+            } while (DateTime.Now <= end);
+            return false;
+        }
+
+        public async void FrameSwitchAttempt(params By[] bys)
```
Keboo: Don't do `async void`. This should be `async Task`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/WebElementExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            int position = element.Location.Y - 200;
+            IJavaScriptExecutor js = (IJavaScriptExecutor)driver;
+            string title = (string)js.ExecuteScript($"window.scrollTo(0,{position})");
+            // Give the browser and javascript a chance to execute.
+            // In the middle of a refactor. Eventually change this out for a centralized wait
+            await Task.Delay(50);
+        }
+
+        public static async void FillInWithAndTab(this IWebElement element, string value)
```
Keboo: Should be `async Task` not `async void`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/WebElementExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿using OpenQA.Selenium;
+using System.Threading.Tasks;
+
+namespace IntelliTect.TestTools.SelenateExtensions
+{
+    public static class WebElementExtensions
+    {
+        public static async void ScrollIntoView(this IWebElement element, IWebDriver driver)
```
Keboo: Should be `async Task` not `async void`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
```diff
@@ -0,0 +1,235 @@
...
+
+            return
+                new ReadOnlyCollection<WebElement>(
+                    WrappedElement.FindElements(by)
+                        .Select(webElement =>
+                        new WebElement(WrappedElement.FindElement(by), by, _Driver))
+                        .ToList());
+        }
+
+        public void Clear()
```
Keboo: Why is this a `void` return method. I would expect that the async refactor would cause this to be:
`public async Task Clear()`.

Similar comment for the other methods in this file.

IntelliTect.TestTools.Selenate/GoogleSearch/ExampleTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        [TestInitialize]
+        public void Setup()
+        {
+            Browser = new GoogleBrowser(BrowserType.Chrome);
+            Google = new GoogleFunctions(Browser);
+            Harness = new GoogleHarness(Browser);
+        }
+
+        [TestMethod]
+        public void SearchForSeleniumOnGoogle()
```
Keboo: I would expect these unit tests to become async methods since the code they are calling should be async methods.
```C#
[TestMethod]
public async Task SearchForSeleniumOnGoogle()
```

PandaMagnus:
I'll drop you a message tomorrow or Tuesday to chat more. I'm hesitant to write this in such a way that the unit test methods are async; it's starting to smell like I might need to handle retries around selenium calls differently. Supporting asynchronous calls would definitely be a nice long term goal to allow some additional state checking if a call is running long, but I'm not sure if that's a good idea for a first pass as most consumers of this will be most familiar with being able to call selenium functions in sequential order.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
```diff
@@ -0,0 +1,274 @@
...
+            return RetryAction("click");
+        }
+
+        // Can this easily be abstracted out to an extension method?
+        // Seems to make more sense as an extension
+        public void Click(int timeToRetry)
+        {
+            TimeToRetry = timeToRetry;
+            Click();
+            TimeToRetry = DefaultRetrySeconds;
```
Keboo: This setting and resetting `TimeToRetry` is a bit problematic. For one, if this value was set to anything other DefautltRetrySeconds, you will have lost that value after calling this method. In general setting and unsetting values like this is problematic. A better design is to pass the `timeToRetry` into the Click method, and then into the `RetryAction` as an optional parameter (using the TimeToRetry if the optional parameter is not set). This avoids needing to keep track of previous values and reset appropriately.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
```diff
@@ -0,0 +1,274 @@
...
+        private readonly IWebDriver _Driver;
+
+        private async Task<string> RetryAction(string action, string text = null)
+        {
+            DateTime end = DateTime.Now.AddSeconds(TimeToRetry);
+            bool reFindElement = !Initialized;
+            List<Exception> retryExceptions = new List<Exception>();
+            while (DateTime.Now <= end)
+            {
+                await Task.Delay(250);
```
Keboo: Again generally you put the delay after the action. That way if it can be completed on the first try the caller isn't forced to wait the 250 milliseconds.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/IntelliTect.TestTools.SelenateExtensions.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningsAsErrors />
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
Keboo: Again this looks like something that should bog in the default `PropertyGroup`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/WebElementExtensions.cs
```diff
@@ -0,0 +1,32 @@
+﻿using OpenQA.Selenium;
+using System.Threading.Tasks;
+
+namespace IntelliTect.TestTools.SelenateExtensions
+{
+    public static class WebElementExtensions
+    {
+        public static void ScrollIntoView(this IWebElement element, IWebDriver driver)
+        {
+            int position = element.Location.Y - 200;
```
Keboo: This 200 seems like it should at least be an optional parameter.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -0,0 +1,251 @@
...
+        /// </summary>
+        /// <param name="func">Function to evaluate</param>
+        /// <param name="secondsToWait">Secondes to wait until timeout / return false</param>
+        /// <returns></returns>
+        public async Task<bool> WaitFor(Func<bool> func, int secondsToWait = 15)
+        {
+            DateTime end = DateTime.Now.AddSeconds(secondsToWait);
+            do
+            {
+                await Task.Delay(500);
```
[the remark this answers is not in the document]
PandaMagnus: Yeah, in this particular case we were trying to eliminate the possibility of Selenium interfering with some JavaScript loading up some data on a webpage. Our thought process was if we pause a second before Selenium is invoked, it would give the JS a chance to finish executing. When I go through and refactor out a common retry method, I'll change the wait to happen after the function invocation.

PR #21, opened 2018-09-22
Curn extracting wait logic

PandaMagnus:


IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
```diff
@@ -0,0 +1,113 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace IntelliTect.TestTools.Selenate
+{
+    public class ConditionalWait
```
Keboo: For these methods, you may consider including overloads that take in the exception types as an array of `Type` parameters. This would allow callers to keep a list of types around in a variable and reuse the list on multiple calls.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
```diff
@@ -0,0 +1,113 @@
...
+            DateTime endTime = DateTime.Now.AddSeconds(seconds);
+            List<Exception> exceptions = new List<Exception>();
+            endTime.AddSeconds(seconds);
+            do
+            {
+                try
+                {
+                    return actionToWaitForComplete();
+                }
+                catch (Exception ex) when (types.ToList().Contains(ex.GetType()))
```
Keboo: The `ToList()` call is not needed here. There is a `.Contains()` LINQ method that will do what you want.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
```diff
@@ -0,0 +1,113 @@
...
+            List<Exception> exceptions = null;
+            endTime.AddSeconds(seconds);
+            do
+            {
+                try
+                {
+                    actionToWaitForComplete();
+                    return;
+                }
+                catch (Exception ex) when (types.ToList().Contains(ex.GetType()))
```
Keboo: The `ToList()` call is not needed here. There is a `.Contains()` LINQ method that will do what you want.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
```diff
@@ -0,0 +1,113 @@
...
+                {
+                    exceptions.Add(ex);
```
Keboo: I don't see where the `exceptions` list is initialized. This method will throw here.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
Keboo: One final naming comment. I would not include "ForSeconds" as part of the method name. That information is already captured with the `seconds` parameter. In addition, because it is an optional parameter, it looks a little weird to see an invocation that has `.WaitForSeconds(() => ...);` but never specifies any seconds. If you want it to be clear to the reader what the second parameter is, consider using named parameters or changing the type to `TimeSpan`.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
```diff
@@ -1,88 +1,118 @@
...
         }
 
         public By By { get; }
-        public bool Initialized => WrappedElement != null;
 
-        // Probably wap out the Task<> properties in favor of Get methods for consistency.
-        // Or scrap this class entirely once retry logic is abstracted out into a single callable method
         string IWebElement.TagName => TagName.ConfigureAwait(false).GetAwaiter().GetResult();
-        public Task<string> TagName => RetryAction("tagname");
+        public Task<string> TagName
```
Keboo: This should really be a method not a property. Properties are expected to be fast quick access to data, not long running async calls.

Same comment from the other properties.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
Keboo: You may also consider making this a static class. Right now most of the code creates a new instance, but this object does not hold any data and merely calls pure methods.

Keboo: I will leave this up to you to decide, but looking at the method signatures it was not immediately apparent to me what each of the generic parameters represented. 
IMO I would rather work with an API more like this:
`await ConditionalWait.WaitFor(() => GetThingy()).Ignore<Exception1, Exception2>();`
Thoughts? If you like it and want help putting something like this together let me know.

IntelliTect.TestTools.Selenate/GoogleSearch/GoogleBrowser.cs
```diff
@@ -8,17 +10,32 @@ public class GoogleBrowser : Browser
...
         {
-            WebElement result = base.FindElement(by);
-            if(!result.Initialized)
+            try
+            {
+                return await base.FindElement(by);
+            }
+            catch (AggregateException ae)
             {
+                var test = ae;
```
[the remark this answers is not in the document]
PandaMagnus: Ah, yep. I was having an issue getting the catch to work. Turns out it was because I had other issues and something other than an AggregateException was being thrown. :) Removed in next commit.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -16,15 +16,31 @@ namespace IntelliTect.TestTools.Selenate
...
+        /// Initializes a Selenium webdriver with some basic settings that work for many websites
+        /// </summary>
+        /// <param name="browser"></param>
         public Browser(BrowserType browser)
         {
             Driver = InitDriver(browser);
         }
 
+        /// <summary>
+        /// Uses an
```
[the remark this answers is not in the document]
PandaMagnus: Added a comment with next commit.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -75,15 +95,10 @@ public IWebDriver InitDriver(BrowserType browser)
         /// <param name="by">Selenium "By" statement to find the element</param>
         /// <param name="secondsToWait">Seconds to wait while retrying before failing</param>
         /// <returns></returns>
-		public WebElement FindElement(By by, int secondsToWait = 5)
+		public Task<IWebElement> FindElement(By by, int secondsToWait = 5)
```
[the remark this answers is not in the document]
PandaMagnus: Is there an easy way to resolve the indentation? I seem to run into it when I've opened the solution in multiple different environments (so I'm assuming the indentation setting is different,) but a ctrl+K+D doesn't seem to solve it. Do I really need to delete and re-indent? :o

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -115,28 +127,28 @@ public IReadOnlyCollection<WebElement> FindElements(By by, int secondsToWait = 5
         /// <returns></returns>
         public async Task<bool> WaitFor(Func<bool> func, int secondsToWait = 15)
         {
-            DateTime end = DateTime.Now.AddSeconds(secondsToWait);
-            do
+            ConditionalWait wait = new ConditionalWait();
```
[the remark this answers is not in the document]
PandaMagnus: Thanks, now being passed into ConditionalWait for every occurrence where it was unused.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -146,102 +158,40 @@ public async Task<bool> WaitFor(Func<bool> func, int secondsToWait = 15)
...
-
-                await Task.Delay(500);
-            }
-            return false;
+            ConditionalWait wait = new ConditionalWait();
+            await wait.WaitForSeconds<NoSuchWindowException>(() => Driver.SwitchTo().Window(title));
         }
 
-        public async Task<IAlert> Alert(int numberOfRetries = 50)
+        public Task<IAlert> Alert(int numberOfRetries = 50)
```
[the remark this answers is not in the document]
PandaMagnus: Removed

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
```diff
@@ -0,0 +1,113 @@
...
+                }
+                await Task.Delay(250);
+            } while (DateTime.Now < endTime);
+            throw new AggregateException(exceptions);
+        }
+
+        private async Task ExecuteWait(Action actionToWaitForComplete, int seconds, params Type[] types)
+        {
+            DateTime endTime = new DateTime();
```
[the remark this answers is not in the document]
PandaMagnus: Oops 😨  Fixed.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
PandaMagnus: Just to make sure I understand: in addition to the methods using generic parameters? Also, in that case I'm going to make the assumption I need to assert that the params list is all Exceptions? That was the original reason for not doing that in the first place, I wasn't sure how to make it fail at compile if someone passed in something not an Exception.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace IntelliTect.TestTools.Selenate
+{
+    public class ConditionalWait
+    {
+        public Task<TResult> WaitForSeconds<T1, TResult>(Func<TResult> func, int seconds = 15)
+            where T1 : Exception
```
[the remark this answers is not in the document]
PandaMagnus: Done

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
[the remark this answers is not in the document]
PandaMagnus: Done

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElement.cs
PandaMagnus: Okay, I was kinda wondering about that. It seemed to be a bit odd/misleading for the property to simply call into a method that's doing a bunch of stuff. I'm considering re-doing this entire class now as it's starting to depart from the default WebElement implementation. I'm not sure what to do about that, as I like implementing IWebElement to maintain some level of compatibility with Selenium, but at the same time the class is really getting beyond what Selenium does.

I'll put this in the task list and think about it, maybe discuss it a bit with some people.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ConditionalWait.cs
PandaMagnus: Went ahead and added additional overloads and checked the list to make sure all arguments are type Exception.

PandaMagnus: I struggled to come up with a good solution to this. I'll skip it for now, but would like to get together with you to chat about it more. TL;DR, I had trouble getting the return type from WaitFor to a final Wait() method.

Keboo: So you might be mixing some concepts, instance vs static methods have no real affect when we are discussing async methods. So no, it will not cause it to get hung up.

Keboo: ah yes happy to sit down with you sometime can go over how to structure the code.

PR #25, opened 2018-10-23
Making conditional wait static

PandaMagnus:


IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/ConditionalWaitTests.cs
```diff
@@ -1,32 +1,177 @@
...
+        {
+            await Assert.ThrowsAsync<AggregateException>(
+                () => Wait.Until<EqualException>(
+                    () => Equals(4, 5), TimeSpan.FromSeconds(1)));
+        }
+
+        [Fact]
+        public async Task CheckForExpectedSuccess()
+        {
+            await Wait.Until<EqualException>(() => Equals(5, 5), TimeSpan.FromSeconds(1));
```
Keboo: This test doesn't contain any expectation, so should probably be renamed to something to indicate that no exception is thrown since "Success" is pretty ambiguous.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/ConditionalWaitTests.cs
```diff
@@ -1,32 +1,177 @@
...
+            Assert.True(sw.Elapsed < TimeSpan.FromSeconds(3), "Timeout was reached when it wasn't expected to be");
+        }
+
+        [Fact]
+        public async Task CheckForTwoExpectedFailuresByParams()
+        {
+            AggregateException ex = await Assert.ThrowsAsync<AggregateException>(
+                () => Wait.Until(() => IterateResults(5, 2), TimeSpan.FromSeconds(2), typeof(InvalidOperationException), typeof(InvalidProgramException)));
+            Assert.NotNull(ex.InnerExceptions.OfType<InvalidOperationException>());
+            Assert.NotNull(ex.InnerExceptions.OfType<InvalidProgramException>());
```
Keboo: Both of these assertions will always pass since the LINQ `OfType` method will simply return an empty enumerable and never null. I think you probably want to do an `Assert.IsTrue` and use a `.Any()` on the end of those.

Similar comment for the tests below as well.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/ConditionalWaitTests.cs
```diff
@@ -1,32 +1,177 @@
...
+
+        [Fact]
+        public async Task CheckFuncParamsForMixedInvalidTypeChecking()
+        {
+            await Assert.ThrowsAsync<ArgumentException>(
+                () => Wait.Until(TestReturnDelegate, TimeSpan.FromSeconds(1), typeof(object), typeof(Exception)));
+        }
+
+        [Fact]
+        public async Task CheckActionParamsForValidTypeChecking()
```
Keboo: It appears this test is all about asserting that `Wait.Until` swallows the passed in exception. The name of the test should reflect that. As a general rule, collapse all of your unit tests and just read the method names. If you can't tell what is being tested without expanding the unit test to look at its contents, consider renaming the method.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/ConditionalWaitTests.cs
```diff
@@ -1,32 +1,177 @@
...
+            AggregateException ex = await Assert.ThrowsAsync<AggregateException>(
+                () => Wait.Until<InvalidOperationException, InvalidProgramException, IndexOutOfRangeException, ArgumentNullException>(() => IterateResults(5, 4), TimeSpan.FromSeconds(2)));
+            Assert.NotNull(ex.InnerExceptions.OfType<InvalidOperationException>());
+            Assert.NotNull(ex.InnerExceptions.OfType<InvalidProgramException>());
+            Assert.NotNull(ex.InnerExceptions.OfType<IndexOutOfRangeException>());
+            Assert.NotNull(ex.InnerExceptions.OfType<ArgumentNullException>());
+        }
+
+        //[Fact]
+        public async Task CheckForMultipleFailuresByGenericOneUnexpected()
```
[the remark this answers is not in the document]
PandaMagnus: Oops... they are, I just forgot to finish them. I'll finish them up for my next commit.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/ConditionalWaitTests.cs
```diff
@@ -39,9 +184,80 @@ private bool TestReturnDelegate()
...
             throw new NullReferenceException();
         }
+
+        private void Equals(int expected, int value)
+        {
+            if (value == expected)
+            {
+                return;
+            }
+            throw new EqualException(expected, value);
```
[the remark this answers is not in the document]
PandaMagnus: XUnit is a little bit backward. The EqualException gets thrown when Assert.Equal fails (i.e., the values you passed in were NOT equal.) However, I'm not opposed to changing this for clarity, since really I just wanted a variety of different exceptions to use to exercise the different paths for catching (or not) exceptions once the wait starts.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/ConditionalWaitTests.cs
```diff
@@ -39,9 +184,80 @@ private bool TestReturnDelegate()
...
+        }
+
+        private void IterateResults(int secondsToFail = 1, int numberOfDifferentExceptions = 1)
+        {
+            if (_End == DateTime.MinValue)
+            {
+                _End = DateTime.Now.AddSeconds(secondsToFail);
+            }
+
+            while (DateTime.Now < _End)
```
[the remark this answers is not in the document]
PandaMagnus: Basically I just needed some sort of method to throw multiple different exceptions. Basically, I want to replicate ui testing behavior, where a framework (be it selenium, WinAppDriver, or etc.) can throw an element not found error, then an element is not interactable error, then etc. until the element hits the expected state. At some point it may pass successfully, but until that point, there may be one of many expected exceptions thrown. Thoughts on if there's a better way to do this?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Wait.cs
```diff
@@ -210,5 +204,13 @@ private async Task ExecuteWait(Action actionToWaitForComplete, TimeSpan timeToWa
             } while (DateTime.Now < endTime);
             throw new AggregateException(exceptions);
         }
+
+        private static void CheckParams(params Type[] exes)
+        {
+            if (!exes.Any(e => e.IsSubclassOf(typeof(Exception)) || e.UnderlyingSystemType == typeof(Exception)))
```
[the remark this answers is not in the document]
PandaMagnus: I totally accidentally did a typeof(e) == typeof(Exception) the first time, which gave me some odd results. :) Apparently the type of a type is a RuntimeType. Fixed.

And thanks for the All tip. :) That is making things work more consistent. ;)

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/ConditionalWaitTests.cs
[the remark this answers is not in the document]
PandaMagnus: Although I did also notice some errors in that method that was causing it to not throw all of the exceptions I was expecting... WHICH because of the issue you pointed out with the Assert.IsNotNull(OfType) things, I never caught. :( I think I at least have it fixed to where things are working as I was intending now.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/ConditionalWaitTests.cs
PandaMagnus: I thought I renamed this last night with my last commit, but apparently not... Unresolving until I can get another commit in.

PandaMagnus: Totally confused. I pull latest and there's no test by this name. It's like the diff tool isn't picking up the change?

Keboo: Looks like you did rename it. It is only showing up in this view because you re-opened the comment thread. It no longer shows up in the branch or the files changed.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/ConditionalWaitTests.cs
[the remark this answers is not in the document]
PandaMagnus: Removed the method in favor of a better method that throws a specified number of exceptions

PandaMagnus:
@Keboo  I think I got everything. The QA side of me got the better of me, and I re-organized my unit tests, which allowed me to better test the difference scenarios for the Wait.Until methods (and of course, that lead to adding several unit tests...) I think I covered everything in your comments.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/BaseTest.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+
+namespace IntelliTect.TestTools.Selenate.Tests
+{
+    public class BaseTest
+    {
+        protected void TestVoidException()
```
Keboo: Consider a more explicit name. Something like `ThrowException`. Then the method below would be `ThrowExceptionWithReturn`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/BaseTest.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            throw new Exception();
+        }
+
+        protected void TestVoidNullRefException()
+        {
+            throw new NullReferenceException();
+        }
+
+        protected void TestReturnNullRefException()
```
Keboo: This is a `void` method. Based on the name and the prior methods I would have expected a non-void method.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/BaseTest.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            ThrowExceptions(secondsToFail, numberOfDifferentExceptions);
+            return true;
+        }
+
+        private void ThrowExceptions(int secondsToFail, int numberOfExceptions)
+        {
+            if (_Timeout == TimeSpan.MinValue)
+            {
+                _Timeout = TimeSpan.FromSeconds(secondsToFail);
```
Keboo: For use inside of this method I would have a local variable `timeOut` rather than setting the field. By using the field you can run into problems if the tests are run in parallel.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/BaseTest.cs
```diff
@@ -0,0 +1,76 @@
...
+                _Sw.Start();
+            }
+
+            while (_Sw.Elapsed <= _Timeout)
+            {
+                _Attempts++;
```
Keboo: This field only appears to be used inside of this method. It should be converted to a local variable.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/BaseTest.cs
PandaMagnus: I'll leave the method as-is for now (XUnit reconstructs for every test, so that should avoid issues at least for now,) but I'll follow-up with you when I get a chance. Basically I just need a different (but known) exception thrown every time this method is invoked up until the timeout is reached.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate.Tests/BaseTest.cs
PandaMagnus: Same as above. I'll follow up with you on a better way to do this.

PR #26, opened 2018-10-25
Changed project settings for package

PandaMagnus:
Setting up project to build nuget packages


PandaMagnus:
Merging in without review since it's solely removing references, adding info for nuget package, and doesn't break any unit tests.

Issue #31, opened 2018-11-14
Break wait out into its own project/package and add variable polling interval


PandaMagnus:
Had a need today to use this on a call WITHOUT Selenium. Basically we have a test that needs to poll a webservice response or database table for a particular value before continuing with the test.

Need to:

- [x] Pull Wait out into own project
~~Add variable polling so we're not hitting the webservice very 250 ms~~
- [x] Package
~~Implement package in test projects~~

PandaMagnus:
Completing the items that are completed with existing PR, striking out the things that are no longer relevant

PR #32, opened 2019-01-08
Using web driver wait

PandaMagnus:
Chris F. found that the expected way of using WebDriverWait is _not_ the way commonly demonstrated online. Swapped out IntelliWait calls for WebDriverWaits with custom exception ignores and delegates passed in. Also sucking in extension project classes as it became clear that the extensions were an integral part of this. In practice, people either wanted it all, or none of it. Extension can be re-broken out later if there's a compelling use case for it.


PandaMagnus:
Changes for Issue #31

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/WebElementExtensions.cs
```diff
@@ -45,9 +46,9 @@ public static void ScrollIntoView(this IWebElement element, IWebDriver driver, i
         /// then tabs out of the field or throws after a certain amount of time
         /// </summary>
         /// <param name="value">Value to send to the element</param>
-        public static async Task FillInWithAndTabWhenReady(this IWebElement element, string value)
+        public static /*async*/ void FillInWithAndTabWhenReady(this IWebElement element, string value)
```
Keboo: I would remove the commented `async`

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/WebElementExtensions.cs
```diff
@@ -58,34 +59,52 @@ public static async Task FillInWithAndTabWhenReady(this IWebElement element, str
...
-                await Wait.Until<NoSuchElementException, 
-                    ElementNotVisibleException, 
-                    ElementNotInteractableException,
-                    StaleElementReferenceException>(
-                    () => element.SendKeys(value), TimeSpan.FromSeconds(secondstoTry));
-                count++;
+                wait.Until( c =>
+                {
+                    element.Clear();
+                    return true;
```
Keboo: Why these return values? `WebDriver.Until` appears to have an `Action` overload so you should not need to have a return value unless you intend to use it somewhere.

IntelliTect.TestTools.Selenate/GoogleSearch/GoogleBrowser.cs
```diff
@@ -10,30 +10,24 @@ public class GoogleBrowser : Browser
...
         }
-        
-        public IWebElement FindElement(By by)
-        {
-            return Find(by).ConfigureAwait(false).GetAwaiter().GetResult();
-        }
 
-        // Instead of doing this, I should just do a separate project demonstrating async usage on something more complex than a Google search
-        public async Task<IWebElement> FindElementAsync(By by)
+        public IWebElement FindElement(By by)
```
kamonson: @PandaMagnus have you considered overriding the By element?

PandaMagnus: I already implemented my own IWebElement once before. It did not turn out well. :laughing:

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElementExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+            element.SendKeys(Keys.Tab);
+        }
+
+        /// <summary>
+        /// Waits for the element to be in a valid state, 
+        /// then clears the current text and uses SendKeys to send the specified value to the element 
+        /// or throws after a certain amount of time
+        /// </summary>
+        /// <param name="value">Value to send to the element</param>
+        public static void FillInWithWhenReady(this IWebElement element, IWebDriver driver, string value, int secondsToTry = 5)
```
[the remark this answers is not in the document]
PandaMagnus: I'll do a bit more research, but primarily I don't want the code to modify Selenium behavior, outside of retrying should the call fail. I did, however, rename the methods to SendKeysWhenReady

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/WebElementExtensions.cs
PandaMagnus: I'm going to revert these changes entirely and leave the project as-is, in preparation for deprecating it. In working with Kyle at a client, we've stumbled on a few key things that make it not make sense to have the extensions in a separate project (at least not these ones.) In fact, a couple of these extensions may even get moved into Browser as functions that you pass the element into.

TL;DR, reverting all changes in this class.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.SelenateExtensions/WebElementExtensions.cs
PandaMagnus: The only one I see is for a Func<IWebDriver, bool> argument?

This is the class that WebDriverWait implements:
https://github.com/SeleniumHQ/selenium/blob/master/dotnet/src/webdriver/Support/DefaultWait%7BT%7D.cs

PR #37, opened 2019-04-05
Moving Selenate solution to root folder

PandaMagnus:


Keboo:
O in that case yes it looks fine

PR #38, opened 2019-05-28
Duplicate types in the same bag fix

kamonson:
When duplicate types are grabbed from the same back there should be no harm in returning the first type of data, this will remove failures thrown by singleOrDefault returning multiple values.


IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -0,0 +1,103 @@
+﻿using Newtonsoft.Json;
```
Keboo: Serious question. Why write your own DI framework when there are other that can be brought in as either source code or dll dependencies? You could still keep this TestBuilder as the base class with your API and hide that it is using a DI container under the hood but it would address the issues you are seeing.

PandaMagnus: I'm struggling with that exact question right now. :( I don't have a good answer. I'm doing some research on the net core implementation now (slowly) to get a better understanding of it to try to help come to a conclusion.

Keboo: That is certainly one option, and it is fairly complex. There are simpler options that can even be brought in through NuGet as straight code files. Like [this one](https://github.com/jonsequitur/PocketContainer). Happy to discuss more if you need.

PandaMagnus: I got a chance to do some reading and tinkering this morning. It looks like AutoFac (if I'm remembering the library name correctly) might be a good option. Microsoft's doesn't natively support property injection, which is problematic for data/UI-heavy tests. 

I'll look into PocketContainer tonight or tomorrow. I really want to get this sorted out before merging to master.

PandaMagnus:
Need resolved by switching to Microsoft.DependencyInjection

PR #44, opened 2019-06-15
Refactoring of ConsoleAssert

MarkMichaelis:
* Added support for error message when equivalency operator did not return true; 
* Refactored execute method; 
* Marked ExpectNoTrimOutput as obsolete as there is an override with normalizeLineEndings; 
* Marked ExpectLike(string expected, char escapeCharacter, Action action) as obsolete as there is a version with default parameters; Implemented CSharpStringEncode() method.


IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -38,6 +41,7 @@ public static string Expect(string expected, Action action, bool normalizeLineEn
         /// <param name="expected">Expected "view" to be seen on the console,
         /// including both input and output</param>
         /// <param name="action">Method to be run</param>
+		[Obsolete]
```
Keboo: Odd indentation. Also Obsolete attribute should include a string indicating how people should alter their code.

IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -137,12 +143,14 @@ private static string Expect(string expected, Action action, Func<string, string
         /// </summary>
         /// <param name="expected">Expected "view" to be seen on the console,
         /// including both input and output</param>
+        /// <param name="escapeCharacter"></param>
         /// <param name="action">Method to be run</param>
-        public static string ExpectLike(string expected, Action action)
+		[Obsolete]
```
Keboo: Odd indentation. Also Obsolete attribute should include a string indicating how people should alter their code.

IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -137,12 +143,14 @@ private static string Expect(string expected, Action action, Func<string, string
...
+        public static string ExpectLike(string expected, char escapeCharacter, Action action)
```
Keboo: Why did you change the method signature (arguably making it a new API), then mark it as obsolete?

IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -251,37 +268,16 @@ public static string Execute(string givenInput, Action action, bool normalizeLin
             }
         }
 
-        private static string GetMessageText(string expectedOutput, string output)
+
+        private static string GetMessageText(string expectedOutput, string output, string equivalentOperatorErrorMessage=null)
```
Keboo: The default value for `equivalentOperatorErrorMessage` is not needed since it appears to be specified at all call sites.

```suggestion
        private static string GetMessageText(string expectedOutput, string output, string equivalentOperatorErrorMessage)
```

IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -351,7 +343,7 @@ private static string CSharpStringEncode(string text)
         /// What a user would see in the console, but with input/output tokens.
         /// </param>
         /// <returns>[0] Input, and [1] Output</returns>
-        private static string[] Parse(string view)  // TODO: Return Tuple instead.
+        private static (string Input,string Output) Parse(string view)  // TODO: Return Tuple instead.
```
Keboo:
```suggestion
        private static (string Input, string Output) Parse(string view) 
```

PR #46, opened 2019-06-24
Common wait implementations

PandaMagnus:


IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+                typeof(ElementNotVisibleException),
+                typeof(StaleElementReferenceException),
+                typeof(NoSuchElementException));
+
+            try
+            {
+                return wait.Until(d => element.Displayed);
+            }
+            catch (WebDriverException ex)
+                when (ex.InnerException.GetType() == typeof(ElementNotVisibleException)
```
Keboo:
```suggestion
                when (ex.InnerException is ElementNotVisibleException
```

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+                        || ex.InnerException.GetType() == typeof(NoSuchElementException))
```
Keboo:
```suggestion
                        || ex.InnerException is NoSuchElementException)
```

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+                typeof(ElementNotInteractableException),
+                typeof(StaleElementReferenceException),
+                typeof(NoSuchElementException));
+
+            try
+            {
+                return wait.Until(d => element.Enabled);
+            }
+            catch(WebDriverException ex)
+                when (ex.InnerException.GetType() == typeof(ElementNotInteractableException))
```
Keboo:
```suggestion
                when (ex.InnerException is ElementNotInteractableException)
```

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElementExtensions.cs
```diff
@@ -48,15 +48,44 @@ public static void ScrollIntoView(this IWebElement element, IWebDriver driver, i
             js.ExecuteScript($"window.scrollTo(0,{position})");
         }
 
+        /// <summary>
+        /// Clears the current text, 
+        /// uses SendKeys to send the specified value to the element, 
+        /// </summary>
+        /// <param name="value">Value to send to the element</param>
+        public static void ClearAndSendKeys(this IWebElement element, string value)
```
Keboo: Personally I hate seeing conjunctions in method names as it is often an indication that a method is doing too much. However in this case perhaps consider a different name such as "SendKeysReplace"?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElementExtensions.cs
```diff
@@ -48,15 +48,44 @@ public static void ScrollIntoView(this IWebElement element, IWebDriver driver, i
...
+            }
+        }
+
+        /// <summary>
+        /// Clears the current text, 
+        /// uses SendKeys to send the specified value to the element, 
+        /// then tabs out of the field
+        /// </summary>
+        /// <param name="value">Value to send to the element</param>
+        public static void ClearAndSendKeysAndTab(this IWebElement element, string value)
```
Keboo: This seems like it would be better as an optional boolean parameters on the end of the previous method. Seeing double conjunctions in the method name is a code smell.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElementExtensions.cs
```diff
@@ -48,15 +48,44 @@ public static void ScrollIntoView(this IWebElement element, IWebDriver driver, i
...
+            element.SendKeys(Keys.Tab);
+        }
+
         /// <summary>
         /// Waits for the element to be in a valid state, then clears the current text, 
         /// uses SendKeys to send the specified value to the element, 
         /// then tabs out of the field or throws after a certain amount of time
         /// </summary>
         /// <param name="value">Value to send to the element</param>
+        //[Obsolete("Use ElementHandler.SendKeysWhenReady for a common implementation of a SendKeys wrapped by a WebDriverWait implementation")]
```
Keboo: Commented out obsolete attribute? Did you mean to comment it out? Same comment on the methods below.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElementExtensions.cs
PandaMagnus: Understood. And it may be another case of trying to solve too much. I'll revisit this week and see if it makes sense to keep around since... I mean really, you could do the same thing with the ClearAndSendKeys("some text" + Keys.Tab); We created a helper method specifically because Maximo needs you to tab out of a field before doing some processing on the UI, which... while it's not unheard of, it's not super duper common. So, again, may be that I'm trying to solve too much here.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/WebElementExtensions.cs
PandaMagnus: Yeah... I probably jumped the gun too quickly on obsoleting some things, and I think I've obsoleted something every version for the last three merges. So I commented it out and am intending to push out a beta package to use at a couple of our clients (which, actually, brings up the super important point of: should beta packages ever come out of master? As soon as I typed this response it dawned on me that they probably shouldn't.)

PR #51, opened 2019-07-17
Adding Dispose to browser

### LINKED ISSUE #50: Browser.cs needs to implement IDisposable
Long story short, when using Selenate with TestFramework, Selenium doesn't get cleaned up correctly because, to date, we've cleaned up Selenium ourselves. Using these two together should be easier to cleanup... I.E. implement IDisposable and do a driver.Dispose.

PandaMagnus:
Resolves #50

### DIFF
--- a/IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
+++ b/IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
@@ -29,7 +29,7 @@ namespace IntelliTect.TestTools.Selenate
     /// Wrapper around a <see cref="IWebDriver"/> that provides numerous
     /// utilities for interacting with the underlying driver.
     /// </summary>
-    public class Browser
+    public class Browser : IDisposable
     {
         /// <summary>
         /// Initializes a Selenium webdriver in Chrome with some basic settings that work for many websites
@@ -187,6 +187,14 @@ namespace IntelliTect.TestTools.Selenate
                 IWebElement element = FindElement(by);
                 wait.Until(f => Driver.SwitchTo().Frame(element));
             }
+        }
+
+        /// <summary>
+        /// Disposes the current Selenium Driver
+        /// </summary>
+        public void Dispose()
+        {
+            Driver.Dispose();
         }
 
         /// <summary>
--- a/IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
+++ b/IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
@@ -283,7 +283,7 @@ namespace IntelliTect.TestTools.Selenate
                 Driver?.Dispose();
             }
 
-            Disposed = true;
+            _Disposed = true;
         }
     }
 }
--- a/IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
+++ b/IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
@@ -201,7 +201,7 @@ namespace IntelliTect.TestTools.Selenate
         /// <summary>
         /// Has Selenium already been disposed?
         /// </summary>
-        protected bool Disposed = false;
+        private bool _Disposed;
 
         /// <summary>
         /// Create a new driver for the given browser type.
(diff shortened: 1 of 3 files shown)


PandaMagnus:
I didn't consider what would happen if the class is derived from (which, while I personally haven't hit a use case, I can definitely imagine some scenarios where that could occur). I'll go ahead and update to follow the BaseClass example.

PandaMagnus:
> @Keboo I didn't consider what would happen if the class is derived from (which, while I personally haven't hit a use case, I can definitely imagine some scenarios where that could occur). I'll go ahead and update to follow the BaseClass example.

@Keboo It's not clear to me from the documentation, would implementing my own SafeHandle be a good option here? That seems weird to me, since really I just need to clean up Selenium in a way that common patterns understand (e.g. if they invoke Browser.cs in a using statement, or if they're using a service provider, or etc.)

Keboo:
Was picturing something much simpler like this:

```
protected virtual void Dispose(bool disposing)
{
    if (disposing)
    {
        Driver?.Dispose();
    }
}

public void Dispose()
{
    Dispose(true);
    GC.SuppressFinalize(this);
}
```

Keboo:
Was picturing something much simpler like this:

```
protected virtual void Dispose(bool disposing)
{
    if (disposing)
    {
        Driver?.Dispose();
    }
}

public void Dispose()
{
    Dispose(true);
    GC.SuppressFinalize(this);
}
```

PandaMagnus:
@Keboo Oh dude, that's way simpler. Okay cool, I can deal with that. :) Do you still think it's worth saving off whether it's been disposed or not and just returning if it has?

Keboo:
Yea, you can if you want. The key details is that calling Dispose multiple times should be handled. Exactly how that gets done is an implementation detail.

PandaMagnus:
@Keboo Done and done (I think. :) )

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -189,6 +189,20 @@ public void FrameSwitchAttempt(int secondsToWait = 15, params By[] bys)
...
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Has Selenium already been disposed?
+        /// </summary>
+        protected bool Disposed = false;
```
Keboo: If you are going to track your disposed state this should be a private field.
```suggestion
        private bool _Disposed;
```

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -252,5 +266,24 @@ protected IWebDriver InitDriver(BrowserType browser)
...
             return driver;
         }
+
+        /// <summary>
+        /// Disposes the current Selenium Driver
+        /// </summary>
+        /// <param name="disposing">Did the call come from Dispose()?</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if(Disposed)
```
Keboo:
```suggestion
            if(_Disposed)
```

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -252,5 +266,24 @@ protected IWebDriver InitDriver(BrowserType browser)
...
+            {
+                return;
+            }
+
+            if(disposing)
+            {
+                Driver?.Dispose();
+            }
+
+            Disposed = true;
```
Keboo:
```suggestion
            _Disposed = true;
```

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
```diff
@@ -189,6 +189,20 @@ public void FrameSwitchAttempt(int secondsToWait = 15, params By[] bys)
...
 
+        /// <summary>
+        /// Disposes the current Selenium Driver
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
```
Keboo:
```suggestion
   ~Browser()
   {
      Dispose(false);
   }
```

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Understood, thank you!

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/Browser.cs
PandaMagnus: Oops, missed this when committing your suggestions. Ended up catching it and changing it in a follow-up commit.

PR #60, opened 2019-10-15
Enhance logging

PandaMagnus:
Enhance logging with the following features:
1) Override default logger correctly
2) Add "finally" block support for final log writeouts, screenshots, etc. (Name may not make sense... Also may need to aggregate the exceptions thrown instead of ignoring them.)
3) Refactor to reduce duplicate code for Finally blocks
4) Add Test Case scope for service provider separate from Test Block scope to make sure loggers are cleaned up correctly (probably redundant since Dispose is called at the end? But just in case.)
5) Add a test case name parameter so we can log it if we have it

### DIFF
--- a/IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/IntelliTectTests.cs
+++ b/IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/IntelliTectTests.cs
@@ -25,7 +25,7 @@ namespace ExampleTests
                     IsBodyAvailable = true
                 };
 
-            TestBuilder builder = new TestBuilder("Test1");
+            TestBuilder builder = new TestBuilder();
             builder
                 .AddLogger<NewLogger>()
                 .AddDependencyService<IWebDriver>(new WebDriverFactory("Chrome").Driver)
@@ -38,7 +38,7 @@ namespace ExampleTests
         [Fact]
         public void RegisterMembership()
         {
-            TestBuilder builder = new TestBuilder("RegisterMembership");
+            TestBuilder builder = new TestBuilder();
             builder
                 .AddDependencyService<IWebDriver>(new WebDriverFactory("Chrome").Driver)
                 .AddDependencyService<Harness.IntelliTectWebpage>()
@@ -50,7 +50,7 @@ namespace ExampleTests
         [Fact]
         public void LogIn()
         {
-            TestBuilder builder = new TestBuilder("LogIn");
+            TestBuilder builder = new TestBuilder();
             builder
                 .AddDependencyService<IWebDriver>(new WebDriverFactory("Chrome").Driver)
                 .AddDependencyService<Account>(new AccountFactory().Account)
--- a/IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/IntelliTectTests.cs
+++ b/IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/IntelliTectTests.cs
@@ -25,7 +25,7 @@ namespace ExampleTests
                     IsBodyAvailable = true
                 };
 
-            TestBuilder builder = new TestBuilder();
+            TestBuilder builder = new TestBuilder("Test1");
             builder
                 .AddDependencyService<IWebDriver>(new WebDriverFactory("Chrome").Driver)
                 .AddDependencyService<Harness.IntelliTectWebpage>()
@@ -37,7 +37,7 @@ namespace ExampleTests
         [Fact]
         public void RegisterMembership()
         {
-            TestBuilder builder = new TestBuilder();
+            TestBuilder builder = new TestBuilder("RegisterMembership");
             builder
                 .AddDependencyService<IWebDriver>(new WebDriverFactory("Chrome").Driver)
                 .AddDependencyService<Harness.IntelliTectWebpage>()
@@ -49,7 +49,7 @@ namespace ExampleTests
         [Fact]
         public void LogIn()
         {
-            TestBuilder builder = new TestBuilder();
+            TestBuilder builder = new TestBuilder("LogIn");
             builder
                 .AddDependencyService<IWebDriver>(new WebDriverFactory("Chrome").Driver)
                 .AddDependencyService<Account>(new AccountFactory().Account)
--- a/IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/IntelliTectTests.cs
+++ b/IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/IntelliTectTests.cs
@@ -27,6 +27,7 @@ namespace ExampleTests
 
             TestBuilder builder = new TestBuilder("Test1");
             builder
+                .AddLogger<NewLogger>()
                 .AddDependencyService<IWebDriver>(new WebDriverFactory("Chrome").Driver)
                 .AddDependencyService<Harness.IntelliTectWebpage>()
                 .AddTestBlock<TestBlocks.NavigateToWebsite>()
--- a/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests.cs
+++ b/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests.cs
@@ -62,7 +62,7 @@ namespace IntelliTect.TestTools.TestFramework.Tests
             TestBuilder builder = new TestBuilder();
             builder.AddTestBlock<ExampleTestBlockWithExecuteArg>("Testing", "Testing2");
 
-            Assert.Throws<ArgumentException>(() => builder.ExecuteTestCase());
+            Assert.Throws<InvalidOperationException>(() => builder.ExecuteTestCase());
         }
 
         [Fact]
@@ -178,7 +178,7 @@ namespace IntelliTect.TestTools.TestFramework.Tests
                 .AddDependencyInstance("Testing")
                 .AddTestBlock<ExampleTestBlockWithMultipleExecuteMethods>();
 
-            Assert.Throws<ArgumentException>(() => builder.ExecuteTestCase());
+            Assert.Throws<InvalidOperationException>(() => builder.ExecuteTestCase());
         }
     }
 
--- a/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests.cs
+++ b/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests.cs
@@ -275,6 +275,41 @@ namespace IntelliTect.TestTools.TestFramework.Tests
                 .AddTestBlock<ExampleTestBlockWithExecuteArg>("Testing")
                 .ExecuteTestCase();
         }
+
+        [Fact]
+        public void OverrideTestCaseNameWithConstructor()
+        {
+            TestBuilder builder = new TestBuilder("Testing");
+            builder
+                .AddLogger<ExampleLogger>()
+                .RemoveLogger()
+                .AddTestBlock<ExampleTestBlockWithExecuteArg>("Testing")
+                .ExecuteTestCase();
+        }
+
+        [Fact]
+        public void OverrideTestCaseNameWithMethod()
+        {
+            TestBuilder builder = new TestBuilder();
+            builder
+                .OverrideTestCaseKey()
+                .AddLogger<ExampleLogger>()
+                .RemoveLogger()
+                .AddTestBlock<ExampleTestBlockWithExecuteArg>("Testing")
+                .ExecuteTestCase();
+        }
+
+        [Fact]
+        public void OverrideTestCaseNameWithMethodOverride()
+        {
+            TestBuilder builder = new TestBuilder();
+            builder
+                .OverrideTestCaseKey("Testing")
+                .AddLogger<ExampleLogger>()
+                .RemoveLogger()
+                .AddTestBlock<ExampleTestBlockWithExecuteArg>("Testing")
+                .ExecuteTestCase();
+        }
     }
 
     public class ExampleTestBlockWithExecuteArg : ITestBlock
--- a/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests.cs
+++ b/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests.cs
@@ -62,7 +62,7 @@ namespace IntelliTect.TestTools.TestFramework.Tests
             TestBuilder builder = new TestBuilder();
             builder.AddTestBlock<ExampleTestBlockWithExecuteArg>("Testing", "Testing2");
 
-            Assert.Throws<InvalidOperationException>(() => builder.ExecuteTestCase());
+            Assert.Throws<TestCaseException>(() => builder.ExecuteTestCase());
         }
 
         [Fact]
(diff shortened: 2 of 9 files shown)


IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/IntelliTectTests.cs
```diff
@@ -25,8 +25,9 @@ public void Test1()
                     IsBodyAvailable = true
                 };
 
-            TestBuilder builder = new TestBuilder();
+            TestBuilder builder = new TestBuilder("Test1");
```
Keboo: There is a fancy option here that could be helpful. If you change the `TestBuilder` constructor to look like this:
```C#
public TestBuilder([CallerMemberName]string caller = null) 
{
    //Caller is automatically provide by the compiler
}
```

You can then make this line just: `TestBuilder builder = new TestBuilder();` and the compiler will automatically put in the name the method that you were in when you called this. Alternatively you can override this by specifying your own string.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests.cs
```diff
@@ -178,7 +178,54 @@ public void TestBlockWithMultipleExecuteMethodsThrows()
...
+        // Figure out the right behavior and fix test before moving further with finally blocks
+        [Fact]
+        public void AddFinallyBlockExecutesCorrectlyWithNegativeCondition()
+        {
+            TestBuilder builder = new TestBuilder();
+            builder
+                .AddTestBlock<ExampleTestBlockWithReturn>(true)
+                .AddFinallyBlock<ExampleFinallyBlock>();
+
+            Assert.Throws<TrueException>(() => builder.ExecuteTestCase());
```
Keboo: `TrueException` is also a weird exception to be expecting here.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -65,147 +83,89 @@ public TestBuilder AddDependencyService<T>()
...
+            //Exception testBlockException = null;
+            using (var testCaseScope = serviceProvider.CreateScope())
             {
-                HashSet<object> testBlockResults = new HashSet<object>();
-                foreach (var tb in TestBlocksAndParams)
-                {
-                    logger.Info($"---{tb.TestBlockType}---");
-                    logger.Debug($"Fetching dependencies.");
-                    object testBlockInstance = null;
+                var logger = testCaseScope.ServiceProvider.GetService<ILogger>() ?? new Log();
```
Keboo: Usually what you do is register this as the default logger in your container, and then let callers override it. Hard coding the `new Log()` means it is not easy for someone to remove logging if they wanted to.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -65,147 +83,89 @@ public TestBuilder AddDependencyService<T>()
...
+                        // Might be more concise to have these as out method parameters instead of if statements after every one
+                        // Also these specific ones should not be overwriting TestBlockException
+                        var testBlockInstance = GetTestBlock(testBlockScope, fb.TestBlockType, logger);
+                        //if (TestBlockException != null) break;
 
-                        // Instead of doing this, might be worth extracting the above for loop into a private method and if that fails, then break out of the foreach we're in now
-                        if (testBlockException != null)
-                            break;
+                        SetTestBlockProperties(testBlockScope, ref testBlockInstance, logger);
+                        //if (TestBlockException != null) break;
```
Keboo: Can we remove the commented code?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -65,147 +83,89 @@ public TestBuilder AddDependencyService<T>()
...
             }
+            
 
             serviceProvider.Dispose();
 
-            if (testBlockException != null)
+            if (TestBlockException != null)
             {
-                throw testBlockException;
+                throw TestBlockException;
```
Keboo: This is probably not what you want here as it will smash the call stack on `TestBlockException`. What you want to do is wrap `TestBlockException` in another exception (probably pick a better exception type that this)
```
throw new Exception(TestBlockException);
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -222,7 +182,150 @@ private string GetObjectDataAsJsonString(object obj)
...
+
+        private void SetTestBlockProperties(IServiceScope scope, ref object testBlockInstance, ILogger logger)
+        {
+            // Populate and log all of our properties
+            var properties = testBlockInstance.GetType().GetProperties();
+            foreach (var prop in properties)
+            {
+                if (!prop.CanWrite)
+                {
+                    logger.Debug(TestCaseName, testBlockInstance.GetType().ToString(), $"Unable to set property {prop}. No setter found.");
```
Keboo: A minor thing here, but a better message might be "Skipping property ..." rather than "Unable to set property..."

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -222,7 +182,150 @@ private string GetObjectDataAsJsonString(object obj)
...
+                }
+
+                prop.SetValue(testBlockInstance, propertyValue);
+                logger.Debug(TestCaseName, testBlockInstance.GetType().ToString(), $"Populated property {prop.Name} with data: {GetObjectDataAsJsonString(prop.GetValue(testBlockInstance))}");
+            }
+        }
+
+        private MethodInfo GetExecuteMethod(IServiceScope scope, object testBlockInstance)
+        {
+            List<MethodInfo> methods = testBlockInstance.GetType().GetMethods().Where(m => m.Name.ToLower() == "execute").ToList();
```
Keboo:
```suggestion
            List<MethodInfo> methods = testBlockInstance.GetType().GetMethods().Where(m => m.Name.ToUpperInvariant() == "EXECUTE").ToList();
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -222,7 +182,150 @@ private string GetObjectDataAsJsonString(object obj)
...
+            }
+
+            logger.Info(TestCaseName, testBlockInstance.GetType().ToString(), $"---Test block completed successfully.---");
+        }
+
         private List<(Type TestBlockType, object[] TestBlockParameters)> TestBlocksAndParams { get; } = new List<(Type TestBlockType, object[] TestBlockParameters)>();
+        private List<(Type TestBlockType, object[] TestBlockParameters)> FinallyBlocksAndParams { get; } = new List<(Type TestBlockType, object[] TestBlockParameters)>();
         private IServiceCollection Services { get; } = new ServiceCollection();
+        private HashSet<object> TestBlockResults = new HashSet<object>();
+        private string TestCaseName { get; set; }
```
Keboo:
```suggestion
        private string TestCaseName { get; }
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -222,7 +182,150 @@ private string GetObjectDataAsJsonString(object obj)
...
+                return;
+            }
+
+            logger.Info(TestCaseName, testBlockInstance.GetType().ToString(), $"---Test block completed successfully.---");
+        }
+
         private List<(Type TestBlockType, object[] TestBlockParameters)> TestBlocksAndParams { get; } = new List<(Type TestBlockType, object[] TestBlockParameters)>();
+        private List<(Type TestBlockType, object[] TestBlockParameters)> FinallyBlocksAndParams { get; } = new List<(Type TestBlockType, object[] TestBlockParameters)>();
         private IServiceCollection Services { get; } = new ServiceCollection();
+        private HashSet<object> TestBlockResults = new HashSet<object>();
```
Keboo:
```suggestion
        private HashSet<object> TestBlockResults { get; } = new HashSet<object>();
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -222,7 +182,150 @@ private string GetObjectDataAsJsonString(object obj)
...
+            {
+                // Probably worth moving these logs outside of the foreach so we don't have to duplicate the message
+                logger.Error(TestCaseName, tbType.ToString(), "Unable to find the test block instance OR all dependencies necessary.");
+                TestBlockException = ex;
+                return null;
+            }
+
+        }
+
+        private void SetTestBlockProperties(IServiceScope scope, ref object testBlockInstance, ILogger logger)
```
Keboo: Is there a reason `testBlockInstance` is passed by ref?

PandaMagnus: I wanted to make it super duper clear that the object being passed in was getting modified, even though it wasn't technically necessary. I suppose re-assigning it from the method return (or the implicitness based on the method name?) is a more appropriate way to do it, now that I look back on it?

Keboo: If the method is going to mutate the object, I would say you should be a `void` return and just pass the object normally. There is no way to really indicate this behavior back to the caller beyond the method name. Given that the name is `SetTestBlockProperties` I think it is perfectly clear that this method will be setting properties on the test block that I pass in. In addition this is a private method so the only people who need to understand this behavior are people who work on this class.

PandaMagnus: Fair enough, will do (as soon as I get a second :) )!

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests.cs
PandaMagnus: I could probably tighten up the test / be more clear in the test name. Basically just need to make sure that the ExampleFinallyBlock throws when expected (_although_ after writing this test, I realized that's not the behavior we ultimately want, so I need to revisit.) The TrueException is what xUnit will throw if an Assert.True fails (which in this case is expected, but again, probably actually _not_ the desired behavior.)

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
PandaMagnus: Just to make sure I understand, will the subsequent logger calls all then be null checked (logger?.Info("blah blah blah");) to simply not log at all should the person using the package do that?

Keboo: Yea you would then null check usages of `logger` to make sure they don't cause an exception. If you want the default implementation of `ILogger` to be a `Log` you would set that up in your `Services` pre-emptively. Then in "AddLogger" you would replace that instance.

PandaMagnus: I couldn't find a way to call the AddLogger method without enabling preview features (if the new nullable reference types features works that way for generic type constraints...) so I added a RemoveLogger method for now. Thoughts on that approach until C# 8.0 is out of preview?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
PandaMagnus: Done with next commit.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
PandaMagnus: OH MY WORD, YES THANK YOU. I had forgot to ask someone about that... I remembered that there was some weirdness (to me) with rethrowing exceptions vs. throwing new ones like this, but I spaced dealing with it... Yes, this is perfect, it retains the ability for the tester to click on the stack trace in the test explorer to see where the test *really* failed, which is what I wanted! Thank you! I'll whip up... probably a custom exception for this and get it in today.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
PandaMagnus: Done with next commit

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
PandaMagnus: Ref removed with next commit

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
PandaMagnus: @Keboo I made some additional changes to the ILogger interface... do you want me to abandon this PR and put in a new one? What's going to be easiest for you to review?

PandaMagnus: Also had to add an additional method to override the test case name/key. xUnit uses the constructor as test setup, so when I instantiated the TestBuilder there, the key was ".ctor". Instead of trying to guess the test case name based on that, I added a method that uses the same [CallerMemberNameAttribute] that a person can call from their test. It's convoluted, and maybe you have a better idea, but it seemed the most straightforward to handle that I don't necessarily know how someone might do test setup/teardown.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests.cs
```diff
@@ -178,7 +178,137 @@ public void TestBlockWithMultipleExecuteMethodsThrows()
...
+        {
+            TestBuilder builder = new TestBuilder();
+            builder
+                .RemoveLogger()
+                .AddTestBlock<ExampleTestBlockWithExecuteArg>("Testing")
+                .ExecuteTestCase();
+        }
+
+        [Fact]
+        public void OverridingLoggerTwiceDoesNotThrow()
```
Keboo:
```suggestion
        public void RemovingLoggerTwiceDoesNotThrow()
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/ILogger.cs
```diff
@@ -2,8 +2,14 @@
 {
     public interface ILogger
     {
+        // Probably need to handle this differently
+        // Maybe a constructor sets the test case name?
```
Keboo: I agree with these comments. In general I don't like public setters like this since it puts the burden on the caller to use the object correctly. But for now it works.

PandaMagnus: I think I had the realization last night that trying to create the logger via the DI service was limiting me from handling this "properly" (i.e. a constructor or some other more controlled method where I can hand in a few strings that set the needed properties.) I think for my next iteration I'll modify how the logger is added/overridden/created so I have the control of handing in constructor arguments. I'd rather get it addressed sooner rather than later since we're starting to use this as multiple clients.

PR #61, opened 2019-10-28
Set up CI with Azure Pipelines

kamonson:


azure-pipelines.yml
```diff
@@ -0,0 +1,33 @@
...
+pool:
+  vmImage: 'windows-latest'
+
+variables:
+  solution: '**/*.sln'
+  buildPlatform: 'Any CPU'
+  buildConfiguration: 'Release'
+
+steps:
+- task: NuGetToolInstaller@1
```
Keboo: This actually bit me. This wont actually install the latest version of NuGet. I believe you will get version 4.3.0, not 5.x

PandaMagnus: Used the example provided to specify NuGet 5.x.

Regarding our convo on the other PR, since it seems like MS is eventually going to deprecate Azure Releases for general Pipelines, would it make sense for me to add the NuGet packing and pushing here, or in its own commit after this is merged in?

Keboo: I am not sure what MS is deprecating.... There is work to re-arrange how pipelines are done, but nothing is really going away just getting cleaned up.
This build pipeline should produce a nuget package, and we should have a separate release pipeline that is responsible for pushing it out. You want a separation of concerns between produce and release.

PandaMagnus: I'll rephrase; the banner here: https://docs.microsoft.com/en-us/azure/devops/pipelines/release/?view=azure-devops _implies_ that they might get rid of release pipelines in the future, but it's kind of ambiguous, since the "new" pipelines appear to support multiple stages.

So to rephrase: do you think it's worth using the "classic" Release Pipeline, or stick with using YML in "Azure Pipelines". If that's not confusing... :)

Keboo: I think what that is in reference to, is that they are wanting to move release pipelines to yaml based ones. So in one sense the current pipelines are going away, but the functionality of having a separate release pipeline is not likely to be removed at all.

PandaMagnus: Okay, gotcha. I'll add the step to generate the NuGet package before merging in, but I won't deploy it in this pipeline.

PandaMagnus:
@Keboo Okay, I added a task to pack up the project and push it to Azure Pipelines for later publishing to NuGet. I'm not sure how to do the versioning, and maybe this is implicating I need to adopt a better branching strategy, but I'm trying to find some way to automatically tag new packages as preview versions (unless the thought is just that I put that in the semver notation?) My expectation would be that after the package is turned around and consumed in a test project, understanding that GoogleSearch doesn't adequately cover all of the scenarios, then we'd bump it to the official release.

Thoughts?

Keboo:
Seems fine for now. I am thinking i am going to write up a blog or at least someone on our new design standards page for doing NuGets since I really like what we came up with for the analyzers package.

PR #70, opened 2020-02-14
Update logging for consistency

PandaMagnus:


IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -177,15 +175,24 @@ public void ExecuteTestCase()
...
-            }
 
+                if(TestBlockException == null)
+                {
+                    logger?.Info("Test case finished successfully.");
+                }
+                else
+                {
+                    logger?.Error($"Test case failed: {TestBlockException.Message}");
+                    logger?.Error($"Stacktrace: {TestBlockException.StackTrace}");
```
Keboo: I would recommend just using `TestBlockException.ToString()`. Though what you have covers most exceptions reasonably some do add additional properties with information.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -278,28 +280,31 @@ private object[] GatherTestBlockArguments(IServiceScope scope, MethodInfo execut
...
+                        object foundResult = null;
+                        try
+                        {
+                            foundResult =
                             TestBlockResults.FirstOrDefault(tbr => tbr.GetType() == executeParams[i].ParameterType)
-                            ?? scope.ServiceProvider.GetService(executeParams[i].ParameterType);
-                        if (foundResult == null)
+                            ?? scope.ServiceProvider.GetRequiredService(executeParams[i].ParameterType);
+                        }
+                        catch(InvalidOperationException ex)
```
Keboo: Why this exception type?

PandaMagnus: @Keboo That's the type that's thrown if GetRequiredService fails to find the asked for type. If this is too vague, I could use an exception filter to see if there are any details more specific to that scenario?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
PandaMagnus: Done and done.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
Keboo: Ah, I would say to just call GetService instead then and then you don't need to catch an exception.

PandaMagnus: I was doing that, but I realized that there's not enough information to continue with the test at this point... so I _was_ doing that, but then immediately threw a new InvalidOperationException... this seemed like a more streamlined way to do it and keep the stacktrace informative. :)

Keboo: The only drawback to letting it throw and catch is that is extra overhead that is not needed. [GetRequiredService](https://source.dot.net/#Microsoft.Extensions.DependencyInjection.Abstractions/ServiceProviderServiceExtensions.cs,56) just does a null check on the return of GetService. I would propose that you do a similar null check, and simply assign TestBlockException = new InvalidOperationException("Useful stuff here").

PandaMagnus: @Keboo Would that apply to the other instances of using GetRequiredService, too? I'd be hesitant to change one and add specific detailing here, but not other places (BUT I'm not opposed to changing. Just want to make sure I understand the purpose and apply it everywhere it's relevant. :) )

Keboo: I would say it should apply everywhere, you are relying on catching that exception. Exceptions are expensive to instantiate and handle, so if you can avoid trowing them and instead handle the case that is better.

PandaMagnus: @Keboo Okay, changes pushed. I changed all cases of GetRequiredService to GetService, then null check the response and assign a new InvalidOperationException(UsefulDetailsHere) to TestBlockException if null.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -278,28 +272,27 @@ private object[] GatherTestBlockArguments(IServiceScope scope, MethodInfo execut
...
-                        // Might be worth changing this to GetRequiredService and wrapping in a try/catch instead of checking if foundResult is null
-                        object foundResult =
-                            TestBlockResults.FirstOrDefault(tbr => tbr.GetType() == executeParams[i].ParameterType)
+                        object foundResult = TestBlockResults.FirstOrDefault(tbr => tbr.GetType() == executeParams[i].ParameterType)
                             ?? scope.ServiceProvider.GetService(executeParams[i].ParameterType);
-                        if (foundResult == null)
+                        if(foundResult == null)
                         {
-                            TestBlockException = new InvalidOperationException("Unable to resolve Execute method arguments");
+                            TestBlockException = new InvalidOperationException($"Unable to find an object or service for Execute parameter {executeParams[i]} on test block {tb.TestBlockType.FullName}.");
```
Keboo:
```suggestion
                            TestBlockException = new InvalidOperationException($"Unable to find an object or service for Execute parameter {executeParams[i].ParameterType} on test block {tb.TestBlockType.FullName}.");
```

PandaMagnus: @Keboo I was lazily trying to get both the parameter name and type. :) Is it worth separating these out to something like {executeParams[i].Name} and {executeParams[i].ParameterType.FullName}?

Keboo: O hm.... yea that works, but in general I prefer to be more explicit.

PandaMagnus: Changed both to be more explicit.

PR #72, opened 2020-02-24
Addtl SendKeys fix

PandaMagnus:
Found a situation where a website is clearing the values in forms sent to it immediately after having the values input (but only the first time; something is probably loading in the background.) Wait for 250ms before checking for the correct value to capture this.


PandaMagnus:
Been using this fix at a client for the last week. Going to get this pushed in since it's a fairly benign change.

PR #74, opened 2020-04-24
Add option to strip VT100 characters

COsborn2:
- Add optional parameter to strip VT100 characters. This will ignore VT100 characters from input and expected output.

### DIFF
--- a/IntelliTect.TestTools.Console.Tests/ConsoleAssertTests.cs
+++ b/IntelliTect.TestTools.Console.Tests/ConsoleAssertTests.cs
@@ -52,17 +52,20 @@ namespace IntelliTect.TestTools.Console.Tests
         }
 
         [TestMethod]
-        [DataRow(@"[49mMontoya", "Montoya")]
-        [DataRow("Inigo[49mMontoya", "InigoMontoya")]
-        [DataRow("Inigo[49m", "Inigo")]
-        [DataRow("[49m", "")]
-        [DataRow("[101mMontoya", "Montoya")]
-        public void ConsoleTester_StringWithVT100Characters_VT100Stripped(string input, string expected)
+        [DataRow("\u001b[49mMontoya", "Montoya", true)]
+        [DataRow("Inigo\u001b[49mMontoya", "InigoMontoya", true)]
+        [DataRow("Inigo\u001b[49m", "Inigo", true)]
+        [DataRow("\u001b[49m", "", true)]
+        [DataRow("\u001b[101mMontoya", "Montoya", true)]
+        [DataRow("\u001b[101mMontoya", "\u001b[101mMontoya", false)]
+        public void ConsoleTester_StringWithVT100Characters_VT100Stripped(string input, 
+            string expected, 
+            bool stripVT100)
         {
             ConsoleAssert.Expect(expected, () =>
             {
                 System.Console.WriteLine(input);
-            }, stripVT100: true);
+            }, stripVT100: stripVT100);
         }
 
         [TestMethod]
--- a/IntelliTect.TestTools.Console.Tests/ConsoleAssertTests.cs
+++ b/IntelliTect.TestTools.Console.Tests/ConsoleAssertTests.cs
@@ -26,7 +26,7 @@ namespace IntelliTect.TestTools.Console.Tests
                 string lname = System.Console.ReadLine();
 
                 System.Console.Write("Hello, {0} {1}.", fname, lname);
-            });
+            }, NormalizeOptions.None);
         }
 
         [TestMethod]
@@ -37,7 +37,7 @@ namespace IntelliTect.TestTools.Console.Tests
             ConsoleAssert.Expect(view, () =>
             {
                 System.Console.Write("Hello World");
-            });
+            }, NormalizeOptions.None);
         }
 
         [TestMethod]
@@ -48,7 +48,7 @@ namespace IntelliTect.TestTools.Console.Tests
             ConsoleAssert.Expect(view, () =>
             {
                 System.Console.WriteLine("Hello World");
-            }, true);
+            }, NormalizeOptions.NormalizeLineEndings);
         }
 
         [TestMethod]
@@ -65,7 +65,7 @@ namespace IntelliTect.TestTools.Console.Tests
             ConsoleAssert.Expect(expected, () =>
             {
                 System.Console.WriteLine(input);
-            }, stripVT100: stripVT100);
+            }, NormalizeOptions.StripVt100 | NormalizeOptions.NormalizeLineEndings);
         }
 
         [TestMethod]
@@ -79,7 +79,7 @@ End";
                 System.Console.WriteLine("Begin");
                 System.Console.WriteLine("Middle");
                 System.Console.WriteLine("End");
-            }, true);
+            }, NormalizeOptions.NormalizeLineEndings);
         }
 
         [TestMethod]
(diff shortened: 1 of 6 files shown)


IntelliTect.TestTools.Console.Tests/ConsoleAssertTests.cs
```diff
@@ -51,6 +51,20 @@ public void ConsoleTester_HelloWorld_TrimLF()
             }, true);
         }
 
+        [TestMethod]
+        [DataRow(@"[49mMontoya", "Montoya")]
```
Keboo, line 55: Rather than putting in the unprintable escape character in the code, I would use the `\u001b` (or a constant) to be clear what that character is. There are lots of unprintable characters that will render as the box

IntelliTect.TestTools.Console.Tests/ConsoleAssertTests.cs
```diff
@@ -51,6 +51,20 @@ public void ConsoleTester_HelloWorld_TrimLF()
...
+        [DataRow("Inigo[49mMontoya", "InigoMontoya")]
+        [DataRow("Inigo[49m", "Inigo")]
+        [DataRow("[49m", "")]
+        [DataRow("[101mMontoya", "Montoya")]
+        public void ConsoleTester_StringWithVT100Characters_VT100Stripped(string input, string expected)
```
Keboo, line 60: I would add in a second test that ensures that if you pass false for `stripVT100` that they are left in.

IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -24,7 +23,9 @@ public static class ConsoleAssert
         /// including both input and output</param>
         /// <param name="action">Method to be run</param>
         /// <param name="normalizeLineEndings">Whether differences in line ending styles should be ignored.</param>
-        public static string Expect(string expected, Action action, bool normalizeLineEndings = true)
+        /// <param name="stripVT100">Whether VT100 color code characters should be ignored.</param>
+        public static string Expect(string expected, Action action, bool normalizeLineEndings = true, 
+            bool stripVT100 = true)
```
MarkMichaelis, line 28: Given we now have normalizeLineEndings and stripVT100, I suspect we should use flags instead of Booleans?  This improves the readability of the calling code in the non-default scenarios as well.

COsborn2: I was trying to add this option without breaking changes to prior versions. I agree some kind of flag would make more sense, but I think that would require changes to the signature and thus breaking changes everywhere?

MarkMichaelis: I'm not sure what that would be true necessarily.
1. We can remove the default from the `normalizeLineEndings` parameter.
2. Add the `NormalizeOptionsFlag` (not certain on the name)? and default it to `NormalizeLineEndings` on and `StripVT100` on.
3. In the method with no options (either Boolean or flags), invoke the method with the `NormalizeOptionsFlag` parameter.

COsborn2: I talked to Kevin. I'll provide a flags overload, that allows the stripVT100 option

IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -42,9 +66,40 @@ public static string Expect(string expected, Action action, bool normalizeLineEn
...
+        public static string Expect(string expected, 
+            Action<string[]> action, 
+            bool normalizeLineEndings = true, 
+            params string[] args)
         {
-            return Expect(expected, ()=>action(args), (left, right) => left == right, normalizeLineEndings);
+            return Expect(expected, 
+                ()=>action(args), 
+                (left, right) => left == right, 
+                normalizeLineEndings ? NormalizeOptions.NormalizeLineEndings : NormalizeOptions.None);
```
MarkMichaelis, line 78: Regardless of whether `normalizeLineEndings` is true or false, shouldn't `NormalizeOptions.StripAnsiEscapeCodes` be set (rather than `NormalizeOptions.None`)?
In fact, consider a `NormalizeOptions.Default` enum value where `NormalizeOptions.Default = NormalizeOptions.NormalizeLineEndings | NormalizeOptions.StripAnsiEscapeCodes`.

IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -42,9 +66,40 @@ public static string Expect(string expected, Action action, bool normalizeLineEn
...
+        }
+        
+        /// <summary>
+        /// <para>
+        /// Performs a unit test on a console-based method. A "view" of
```
MarkMichaelis, line 83: Glad to see the help text!!

IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -67,14 +122,14 @@ public static string Expect(string expected, Action<string[]> action, bool norma
         public static void Expect<T>(string expected, Func<string[], T> func, T expectedReturn = default, params string[] args)
         {
             T @return = default;
-            Expect(expected, () => @return = func(args));
+            Expect(expected, () => @return = func(args), NormalizeOptions.None);
```
MarkMichaelis, line 125: Why default to `NormalizeOptions.None`?  Shouldn't we have both normalize line endings and strip ansi escape code to true?  (In other words `NormalizeOptions.Default` once we have it?)

IntelliTect.TestTools.Console/ConsoleAssert.cs
```diff
@@ -125,7 +182,7 @@ public static void Expect<T>(string expected, Func<T> func, T expectedReturn)
         /// <param name="func">Method to be run</param>
         /// <param name="args">Args to pass to the function.</param>
         public static void Expect(string expected, Action<string[]> func, params string[] args) =>
-            Expect(expected, () => func(args));
+            Expect(expected, () => func(args), NormalizeOptions.NormalizeLineEndings);
```
MarkMichaelis, line 185: Why not just rely on the parameter default - which means `NormalizeOptions.StripAnsiEscapeCodes` would be se.

IntelliTect.TestTools.Console/NormalizeOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+        
+        /// <summary>
+        /// Whether differences in line ending styles should be ignored.
+        /// </summary>
+        NormalizeLineEndings = 1,
+        
+        /// <summary>
+        /// Whether ansi color code characters should be ignored.
+        /// </summary>
+        StripAnsiEscapeCodes = 2,
```
MarkMichaelis, line 24: How about a `Default` value set to `StripAnsiEscapeCodes | NormalizeLineEndings`?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/IntelliTect.TestTools.TestFramework.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
MarkMichaelis, line 4: Thanks for upgrading this!

IntelliTect.TestTools.Console/ConsoleAssert.cs
COsborn2: Yes! Thank you

PR #78, opened 2020-05-07
Added support for running Execute() tests on Linux.

MarkMichaelis:


MarkMichaelis:
@COsborn2 - can you please check this works on Mac and Linux. If not, suggestions welcome. :)

COsborn2:
@MarkMichaelis "ping.exe" won't exist on UNIX platforms, so that doesn't work. I'll try to think of something that exists on a UNIX platform and on Windows so we can test that method.

MarkMichaelis:
Oh.. bummer... I thought `ping` would work.  I tried it on WSL - which is how I came up with the different command-line arguments.  Hmmm.....

IntelliTect.TestTools.Console.Tests/ConsoleAssertTests.cs
```diff
@@ -91,29 +91,21 @@ public void ConsoleTester_HelloWorld_MissingNewline()
...
-Pinging * ?::1? with 32 bytes of data:
-Reply from ::1: time*",
-                "ping.exe", "-n 4 localhost", out string standardOutput, out _);
-                Assert.IsTrue(standardOutput.ToLower().IsLike($"*{ Environment.MachineName.ToLower()}*"));
+                pingArgs = pingArgs.Replace("-c ", "-n ");
             }
+
+            ConsoleAssert.ExecuteProcess(
+                expected,
+                "ping.exe", pingArgs, out string standardOutput, out _);
```
Keboo, line 107: The problem here, is just because a tool exists on one linux distribution, doesn't guarantee that it will exist on any others. For example, Alpine linux (popular with many container like things) is pretty bare bones and has almost nothing by default. While something like ubuntu likely has it by default. Testing in WSL is fine, but you have to remember that they put in a lot of magic to forward to windows commands (which is why I suspect it worked for you).

As for making this particular example work (if we assume the system has `ping` installed). I suspect if you just drop the ".exe" portion it will likely work.

MarkMichaelis:
Yeah... I was surprised I had the "exe" extension and realized it was a problem when Cameron pointed raised the issue. :(
What if we check for "ping" and `Assert.Inconclusive()` if it isn't available? Of course I'm not sure of an easy way to do that besides attempting to execute the process and catching the exception.
Do you have a better suggestion that "ping?" Preferably we use something that is available on both Linux and Windows in Azure DevOps.

BenjaminMichaelis:
Addressed in https://github.com/IntelliTect/TestTools.Console/pull/1

PR #79, opened 2020-06-05
Add IntelliTect.TestTools.Data to make DB testing easier

COsborn2:
- Add DbFixture that takes DbContext and setups up in memory Database Connections
- Restrict Database access to Func, to ensure that invokers access new DbContext to avoid caching

### DIFF
--- a/IntelliTect.TestTools.Data.Test/BadDbContext.cs
+++ b/IntelliTect.TestTools.Data.Test/BadDbContext.cs
@@ -1,0 +1,13 @@
+﻿using Microsoft.EntityFrameworkCore;
+
+namespace IntelliTect.TestTools.Data.Test
+{
+    public class BadDbContext : DbContext
+    {
+        public DbSet<Person> Persons { get; set; }
+
+        public BadDbContext()
+        {
+        }
+    }
+}
--- a/IntelliTect.TestTools.Data.Test/BadDbContext.cs
+++ b/IntelliTect.TestTools.Data.Test/BadDbContext.cs
@@ -1,13 +1,19 @@
-﻿using Microsoft.EntityFrameworkCore;
+﻿using System;
+using Microsoft.EntityFrameworkCore;
 
 namespace IntelliTect.TestTools.Data.Test
 {
     public class BadDbContext : DbContext
     {
-        public DbSet<Person> Persons { get; set; }
+        private object MyDependency { get; }
 
         public BadDbContext()
         {
+        }
+
+        public BadDbContext(DbContextOptions options, object someOtherDependency) : base(options)
+        {
+            MyDependency = someOtherDependency ?? throw new ArgumentNullException(nameof(someOtherDependency));
         }
     }
 }
--- a/IntelliTect.TestTools.Data.Test/DatabaseFixture.SampleDbContext.Tests.cs
+++ b/IntelliTect.TestTools.Data.Test/DatabaseFixture.SampleDbContext.Tests.cs
@@ -1,0 +1,58 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using IntelliTect.IntelliTime.Data.Test.Util;
+using Xunit;
+
+namespace IntelliTect.TestTools.Data.Test
+{
+    public class DatabaseFixtureSampleDbContextTests : IClassFixture<DatabaseFixture<SampleDbContext>>
+    {
+        private readonly DatabaseFixture<SampleDbContext> _DatabaseFixture;
+
+        public DatabaseFixtureSampleDbContextTests(DatabaseFixture<SampleDbContext> databaseFixture)
+        {
+            _DatabaseFixture = databaseFixture ?? throw new ArgumentNullException(nameof(databaseFixture));
+        }
+
+        [Fact]
+        public void DatabaseFixture_CanPerformDatabaseOperation()
+        {
+            var person = FakesFactory.Create<Person>();
+
+            _DatabaseFixture.PerformDatabaseOperation(async context =>
+            {
+                await context.Persons.AddAsync(person);
+                await context.SaveChangesAsync();
+            });
+
+            _DatabaseFixture.PerformDatabaseOperation(context =>
+            {
+                var personInDatabase = context.Persons.SingleOrDefault();
+
+                Assert.NotNull(personInDatabase);
+                Assert.Equal(person.Age, personInDatabase.Age);
+                Assert.Equal(person.Name, personInDatabase.Name);
+
+                return Task.CompletedTask;
+            });
+        }
+
+        [Fact]
+        public void DatabaseFixture_EFInMemoryLoggingEnabled_LogsPopulated()
+        {
+            var person = FakesFactory.Create<Person>();
+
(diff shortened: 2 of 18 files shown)


IntelliTect.TestTools.Data.Test/BadDbContext.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.EntityFrameworkCore;
+
+namespace IntelliTect.TestTools.Data.Test
+{
+    public class BadDbContext : DbContext
+    {
+        public DbSet<Person> Persons { get; set; }
+
+        public BadDbContext()
```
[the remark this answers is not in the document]
COsborn2: I should probably add additional constructors that dont match the required signature. One of the test classes is reflecting on this type, and throwing when the expected ctor signature is not found.

IntelliTect.TestTools.Data/Log.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using Microsoft.Extensions.Logging;
+
+namespace IntelliTect.TestTools.Data
+{
+    public class Log
+    {
+        private readonly LogLevel _LogLevel;
```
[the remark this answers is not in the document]
COsborn2: ...yes... they should all be public... Thanks

IntelliTect.TestTools.Data/DatabaseFixture.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            if (!alreadyCreated)
+            {
+                db.Database.EnsureCreated();
+            }
+
+            return db;
+        }
+
+        public async Task PerformDatabaseOperation(Func<TDbContext, Task> operation)
```
Steborino, line 80: Perhaps add method documentation specifically calling out the intended use of this method — i.e. to prevent caching, so _not_ around an entire test method.

Steborino, line 80: Mentioned this idea to @COsborn2, who said I should post it in case you have any thoughts, @Keboo:

> I did have one idea, which is that you could further enforce the paradigm by creating a method where you have to pass in a callback for each segment of your test (arrange/act/assert) and the context is refreshed between each. But then this felt too clunky and prescriptive.

Keboo: I assume we are talking about applying [xml docs](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/) to the public API. I think that is a great idea, especially if the goal is to distribute as a NuGet (source link would be the other good bit to add as well). In general people want to see both when interacting with a library.

COsborn2: There should definitely be documentation on all public members of this API. Totally dropped the ball on that. Thanks!

COsborn2: FWIW the Directory.Build.Props should add sourcelink

Steborino: @Keboo - I read your comment and assumed you responded to my other comment (albeit on the same line) — I agree with everything you just said. 👍 

But my intent was to + you regarding a different idea, which I quoted in the parent comment — specifically, a stricter API that could prevent users from shooting themselves in the foot (i.e. only calling `PerformDatabaseOperation` once in a test method). My original idea is too restrictive, I think, but perhaps you might have a better idea.

Keboo: Ah sorry, yes @COsborn2 pointed out that I messed that one up. I am not sure there is really going to be an API solution to that problem. Though in nearly all cases the various portions of the test are easily separate, the difference between each section is, at most, reflected with a comment. Because even with a restrictive API that tries to force you to pass three separate callback delegates, aside from a CR, there is nothing stopping someone from putting the entire method in a single callback. Perhaps the better solution is a rename to make it more obvious how the method is intended to be used. Just as an idea (not sure I like it), perhaps something like `WithContext` that way if someone puts their entire unit test method inside of the callback it is clear that the entire test is acting on the same context. Thoughts?

COsborn2: Added XML comments, hopefully should make things more obvious.

IntelliTect.TestTools.Data/DatabaseFixture.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        /// <summary>
+        /// If <see cref="InMemoryLogger"/> is configured and DbContext has been accessed at least once, returns a
+        /// Dictionary of all InMemoryLoggers
+        /// </summary>
+        /// <returns>Dictionary with category name, and instance of all InMemoryLoggers</returns>
+        /// <exception cref="InvalidOperationException">InMemoryLogger is not configured, or DbContext has not yet
+        /// been accessed</exception>
+        public ConcurrentDictionary<string,InMemoryLogger> GetInMemoryLoggers()
```
Keboo, line 104:
```suggestion
        public ConcurrentDictionary<string, InMemoryLogger> GetInMemoryLoggers()
```

IntelliTect.TestTools.Data/DatabaseFixture.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            if (ServiceProvider is null)
+            {
+                throw new InvalidOperationException("ServiceCollection is not yet initialized. " +
+                                                    "Perform some database operation to initialize loggers");
+            }
+
+            if (!(ServiceProvider.GetService<ILoggerProvider>() is InMemoryLoggerProvider loggerProvider))
+            {
+                throw new InvalidOperationException($"{typeof(ILoggerProvider)} of type " +
```
Keboo, line 114: Did you mean to use `typeof` to get the full name (in which case I would recommend being explicit and calling the FullName property), or did you mean to use `nameof` here?

COsborn2: I think I did mean to use typeof. Whats the difference between letting it call ToString implicitly vs full name property? Is this a convention?

Keboo: One is assumed behavior that (though very unlikely) could change, the other is explicit.

IntelliTect.TestTools.Data.Test/DatabaseFixture.SampleDbContext.Tests.cs
```diff
@@ -0,0 +1,145 @@
...
+        [Fact]
+        public async Task DatabaseFixture_WriteAndRetrieveObjectWithNoCaching_NoIncludedNavigationProperties()
+        {
+            var person = FakesFactory.Create<Person>();
+
+            await _DatabaseFixture.PerformDatabaseOperation(async context =>
+            {
+                await context.Persons.AddAsync(person);
+                await context.SaveChangesAsync();
+            });
```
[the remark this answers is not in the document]
COsborn2: Definitely. I was trying to be explicit in these tests, but as these expand that would definitely be an improvement. Thanks

PR #83, opened 2020-06-17
Testframework analyzers

PandaMagnus:
Add analyzers and resolve warnings and errors

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,4 @@
 ﻿[*.cs]
 
-# CA1822: Mark members as static
-dotnet_diagnostic.CA1822.severity = silent
-
 # CA1303: Do not pass literals as localized parameters
-dotnet_diagnostic.CA1303.severity = silent
+dotnet_diagnostic.CA1303.severity = warning
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,1 @@
 ﻿[*.cs]
-
-# CA1303: Do not pass literals as localized parameters
-dotnet_diagnostic.CA1303.severity = warning
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+﻿[*.cs]
+
+# CA1822: Mark members as static
+dotnet_diagnostic.CA1822.severity = silent
+
+# CA1303: Do not pass literals as localized parameters
+dotnet_diagnostic.CA1303.severity = silent
--- a/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/IntelliTect.TestTools.TestFramework.Tests.csproj
+++ b/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/IntelliTect.TestTools.TestFramework.Tests.csproj
@@ -1,9 +1,13 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
 
     <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>CA1822, CA1303</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
--- a/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/IntelliTect.TestTools.TestFramework.Tests.csproj
+++ b/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/IntelliTect.TestTools.TestFramework.Tests.csproj
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="IntelliTect.Analyzers" Version="0.1.6" />
     <PackageReference Include="xunit" Version="2.3.1" />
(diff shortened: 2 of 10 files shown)


.editorconfig
```diff
@@ -0,0 +1,7 @@
+﻿[*.cs]
+
+# CA1822: Mark members as static
+dotnet_diagnostic.CA1822.severity = silent
+
+# CA1303: Do not pass literals as localized parameters
+dotnet_diagnostic.CA1303.severity = silent
```
PandaMagnus, line 7: I was trying to do this just for the unit test project. Did I do that right?

Keboo: This will affect everything in the repository. You probably want to supress the warning with NoWarn in the csproj

PandaMagnus: @Keboo I'm torn on 1303... As probably expected, when I removed these, some 1303 error showed up in the main project that I'm trying real hard not to turn off any of the analyzers on. But when I looked up the recommended fix, it seems... a bit excessive?

https://docs.microsoft.com/en-us/dotnet/api/system.resources.resourcemanager?view=netcore-3.1

Suggestions on following the approach recommend in the URL (creating a resource manager class backed by a .resources file)?

Keboo: That sems excessive unless we plan to support localization. I think simply disabling it for your project is the simplest option for now.

PandaMagnus: @Keboo  Totally forgot to come back to this. Moved to NoWarns in each project and reverted .editorconfig settings.

PandaMagnus: @Keboo  It looks like the NoWarn tags I added aren't getting recognized by the build server/pipeline? That seems strange... I'm gonna try a couple of different things, but may require your assistance if I can't get it figured out.

PandaMagnus: @Keboo Details, details... I forgot to add the NoWarn for Release builds. :)

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/IntelliTect.TestTools.TestFramework.Tests.csproj
```diff
@@ -1,11 +1,23 @@
...
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningLevel>4</WarningLevel>
+    <NoWarn>1701;1702;CA1822;CA1303</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
Keboo, line 15: Both this PropertyGroup and the one above appear to be identical. Any reason we can't just move these values into the default property group on line 3?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.csproj
```diff
@@ -2,20 +2,22 @@
...
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <WarningsAsErrors>NU1605, NU5105</WarningsAsErrors>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningLevel>4</WarningLevel>
+    <NoWarn>1701;1702;CA1303</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
Keboo, line 13: Same comment here, rather than duplicating this between debug and release build, just put the values inside of the default property group.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -86,6 +86,7 @@ public TestBuilder AddDependencyService<T>()
         /// <returns>This</returns>
         public TestBuilder AddDependencyInstance(object objToAdd)
         {
+            if (objToAdd == null) throw new ArgumentNullException(nameof(objToAdd));
```
Keboo, line 89: nit:
```suggestion
            if (objToAdd is null) throw new ArgumentNullException(nameof(objToAdd));
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/IntelliTect.TestTools.TestFramework.Tests.csproj
PandaMagnus: Honestly? I wasn't sure best practice so erred on the side of verboseness. Moved up to the default property group. Thanks!

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.csproj
PandaMagnus: Same as above. Moved.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
PandaMagnus: Pretty sure you catch this every time I forget to do "is". :) Thanks!

PR #86, opened 2020-06-19
Move from netcore3.1 to netstandard2.0

COsborn2:


IntelliTect.TestTools.Data/DatabaseFixture.cs
```diff
@@ -15,7 +15,7 @@ public class DatabaseFixture<TDbContext> : IDisposable where TDbContext : DbCont
 
         private Lazy<DbContextOptions<TDbContext>> Options { get; }
 
-        private IServiceProvider? ServiceProvider { get; set; }
+        private IServiceProvider ServiceProvider { get; set; }
```
Keboo, line 18: I don't think this is correct. It is not set in the ctor, so it will be null directly after instantiation.

IntelliTect.TestTools.Data/DatabaseFixture.cs
```diff
@@ -32,7 +32,7 @@ public DatabaseFixture()
         /// <summary>
         /// Fired when loggers are being setup. Immediately follows adding the InMemoryLogger
         /// </summary>
-        public event EventHandler<ILoggingBuilder>? BeforeLoggingSetup;
+        public event EventHandler<ILoggingBuilder> BeforeLoggingSetup;
```
Keboo, line 35: Won't this be null directly after instantiation?

IntelliTect.TestTools.Data/Log.cs
```diff
@@ -35,7 +35,7 @@ public class Log
         public Log(
             LogLevel logLevel,
             EventId eventId,
-            object? state,
+            object state,
```
Keboo, line 38: I would add a null check on the reference types since this class is public.

IntelliTect.TestTools.Data/InMemoryLogger.cs
```diff
@@ -43,7 +43,7 @@ public bool IsEnabled(LogLevel logLevel)
         /// </summary>
         public IDisposable BeginScope<TState>(TState state)
         {
-            throw new NotImplementedException();
+            return null;
```
Keboo, line 46: I think you want `return this;` don't you? Otherwise scoped invocations I would expect to fail.

IntelliTect.TestTools.Data/Log.cs
COsborn2: I dont think Log implementations are guaranteed to provide a state. That being said I don't know if this is worthy of public. Do you think I should just make the constructor internal?

IntelliTect.TestTools.Data/InMemoryLogger.cs
COsborn2: It looks like null return is expected, looking at other implementations. https://github.com/horsdal/InMemoryLogger/blob/master/src/InMemoryLogger.cs

PR #90, opened 2020-08-23
Update selenium

### LINKED ISSUE #89: Fix Browser.TakeScreenshot() comment
It specifies the wrong location.

PandaMagnus:
Fixes #89  and also updates to latest Selenium and fixes a change to ICapabilities Selenium made.

### DIFF
--- a/IntelliTect.TestTools.Selenate/GoogleSearch/ExampleTests.cs
+++ b/IntelliTect.TestTools.Selenate/GoogleSearch/ExampleTests.cs
@@ -47,7 +47,7 @@ namespace GoogleSearch
         [TestMethod]
         public void TakeScreenshotSavesFile()
         {
-            string path = Path.Combine("C:", "temp", "screenshots");
+            string path = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
             Directory.Delete(path, true);
 
             Google.SearchForItem("selenium automation");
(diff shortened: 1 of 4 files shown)


IntelliTect.TestTools.Selenate/GoogleSearch/ExampleTests.cs
```diff
@@ -44,17 +44,38 @@ public void ReturnToHomepage()
                 "Search results displayed when they were not expected");
         }
 
-        [TestCleanup]
-        public void Teardown()
+        [TestMethod]
+        public void TakeScreenshotSavesFile()
         {
-            // Two ways to take screenshots.
+            string path = Path.Combine("C:", "temp", "screenshots");
```
Keboo, line 50: Rather than using hard coded paths. It also doesn't guarantee you will have write permissions on the directory. I would suggest leveraging something more like:
```suggestion
            string path = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
```

Though I realize the base method relies on this path, so it might not be a viable option.

PandaMagnus: @Keboo Okay, that's pretty neat. :) I figure C:\temp would be a decent bet for being able to write to. Although since TakeScreenshot() already attempts to write there, I suppose this one I could specifically change to write to the current directory.

PR #102, opened 2020-11-23
Adding net5 support to IntelliTect.TestTools.Data

Keboo:
Removed extra nuget packages from the repo and fixed gitignore


COsborn2 (approved):
Surprised it was that easy. Sweet.

PR #104, opened 2021-01-08
Json serialization

PandaMagnus:
Wrap JSON serialization in try/catch. Certain things in tests (e.g. WebDriver) can't and shouldn't be serialized.

This is a bandaid. May need to come up with a better long-term solution, e.g. an attribute to opt into logging instead of just assuming everything should be logged.


PandaMagnus:
Actually, before I merge this, I'll investigate using ReferenceHandler.Preserve.

PandaMagnus:
> Actually, before I merge this, I'll investigate using ReferenceHandler.Preserve.

No, nevermind. I changed my mind, I'll leave for investigating if doing a try/catch is the right long term solution. I need to get this merged by Monday so one of our clients can start using it (I'll let them know the caveat that there are some things that still need to be worked through. Luckily we won't have full adoption of this for another few weeks.)

PandaMagnus:
Since this is just wrapping a call in a try/catch, I'm going to go ahead and merge this. If anyone is very upset by that, let me know and I'll unlist the package.

PR #105, opened 2021-01-23
Additional AddService and AddInstance methods

PandaMagnus:
At a client, I had to extract some common code out into an internal package, so I used some interface definitions to ensure cross-compatibility. To support that, I need to add a couple additional methods...
- AddDependencyService<TService, TImplemintation>()
- AddDependencyInstance<T>(object objToAdd()

So we can do things like:

AddDependencyService<IOracleConnection, SomeDatabaseClass>().

### DIFF
--- a/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
+++ b/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
@@ -79,9 +79,9 @@ namespace IntelliTect.TestTools.TestFramework
             return this;
         }
 
-        public TestBuilder AddDependencyService<TService,TImplementation>()
+        public TestBuilder AddDependencyService<TServiceType,TImplementationType>()
         {
-            Services.AddScoped(typeof(TService), typeof(TImplementation));
+            Services.AddScoped(typeof(TServiceType), typeof(TImplementationType));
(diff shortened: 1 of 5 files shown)


IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -79,6 +79,12 @@ public TestBuilder AddDependencyService<T>()
             return this;
         }
 
+        public TestBuilder AddDependencyService<TServiceType,TImplementationType>()
+        {
+            Services.AddScoped(typeof(TServiceType), typeof(TImplementationType));
```
Keboo, line 84: Why does this add a Scoped instance while the other `AddDependencyService` methods do singleton?

PandaMagnus: Ah, okay, so it depends on how the item is being added. If it's by type, the assumption (and so far, the need to-date) is that it's something you want managed solely by the DI service within the context of a single test case (hence, scoped). If it's added by object, then it's added a singleton on the assumption that something *outside* the scope of the DI container is managing that object (e.g., xUnit's data collections in order to share an object amongst multiple tests, which xUnit will handle cleaning up.)

PR #106, opened 2021-02-11
Change try/catch around JSON serialization to a general Exception catch

PandaMagnus:
Long story short, I found some different behavior in how System.Text.Json's JsonSerializer.Serialize works. In one specific scenario (attempting to serialize an object results in a property on said object throwing an exception) I saw two different behaviors:
1) Throw a TargetInvocationException with an inner exception of whatever the object threw
2) Throw the original exception (in this case, it happened to be a Selenium WebDriverException)

To prevent having to explicitly catch just all of the exceptions, the code change is to catch a regular ol' Exception in this one specific case, because we don't want logging to behave different and risk stopping a test because some property couldn't be evaluated at log-time.

### DIFF
--- a/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
+++ b/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
@@ -220,7 +220,7 @@ namespace IntelliTect.TestTools.TestFramework
             catch (Exception e)
 #pragma warning restore CA1031 // Do not catch general exception types
             {
-                return $"Unable to serialize object {obj?.GetType()} to JSON. Mark the relevant property with the [JsonIgnore] attribute: {e.Message}";
+                return $"Unable to serialize object {obj?.GetType()} to JSON. Mark the relevant property with the [JsonIgnore] attribute: {e}";
             }
             
         }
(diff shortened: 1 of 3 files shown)


IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/TestBuilder.cs
```diff
@@ -208,12 +208,17 @@ public void ExecuteTestCase()
...
+            // In this one specific case, catch all exceptions and move on to provide standard behavior to all package consumers.
+            // TL;DR: we don't want logging failures to interrupt the test run.
             try
             {
                 return JsonSerializer.Serialize(obj, new JsonSerializerOptions { WriteIndented = true });
             }
-            catch (JsonException e)
+#pragma warning disable CA1031 // Do not catch general exception types
+            catch (Exception e)
+#pragma warning restore CA1031 // Do not catch general exception types
```
Keboo, line 221: This seems fine, however I would advocate not just logging `e.Message` this can be problematic if the useful information is contained inside of its inner exception. I would use `e.ToString()` instead

PandaMagnus: Thank you! I always get those two confused. I'll make that change real quick.

PandaMagnus: Okay, change made. One of the versions I tested did throw an exception with an inner exception, so this should be more helpful. Thanks again!

PR #112, opened 2021-09-27
Add method to use a predicate to find a specific IWebElement. Needed …

PandaMagnus:
Add a helper method to find a single IWebElement in a list using a predicate.
Add unit tests for scenarios.

### DIFF
--- a/IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
+++ b/IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
@@ -93,19 +93,26 @@ namespace IntelliTect.TestTools.Selenate
             }
         }
 
+        /// <summary>
+        /// Checks if any element found by <see cref="Locator"/> matches a predicate.
+        /// </summary>
+        /// <param name="predicate">The criteria to attempt to match on.</param>
+        /// <returns></returns>
         public IWebElement GetSingleExistingElement(Func<IWebElement, bool> predicate)
         {
             IWait<IWebDriver> wait = Wait;
             wait.IgnoreExceptionTypes(typeof(NoSuchElementException));
             return wait.Until(d =>
             {
-                var foundElems = d.FindElements(Locator).Where(predicate).ToList();
-                if(foundElems.Count != 1)
+                var foundElems = d.FindElements(Locator);
+                if (foundElems is null || foundElems.Count == 0) throw new NoSuchElementException($"No element found matching pattern: {Locator}");
+                var foundElem = foundElems.Where(predicate).ToList();
+                if (foundElem.Count != 1)
                 {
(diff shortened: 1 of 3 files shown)


IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
```diff
@@ -91,5 +92,21 @@ public bool ContainsText(string text)
...
+            wait.IgnoreExceptionTypes(typeof(NoSuchElementException));
+            return wait.Until(d =>
+            {
+                var foundElems = d.FindElements(Locator).Where(predicate).ToList();
+                if(foundElems.Count != 1)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(predicate), 
+                        "The search criteria did not return exactly one result.");
+                }
+                return foundElems.Single();
```
Keboo, line 108: nit: it is slightly better perf to use the indexer rather than the LINQ method here since you have already checked the count.
```suggestion
                return foundElems[0];
```

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
```diff
@@ -91,5 +92,21 @@ public bool ContainsText(string text)
...
+        public IWebElement GetSingleExistingElement(Func<IWebElement, bool> predicate)
+        {
+            IWait<IWebDriver> wait = Wait;
+            wait.IgnoreExceptionTypes(typeof(NoSuchElementException));
+            return wait.Until(d =>
+            {
+                var foundElems = d.FindElements(Locator).Where(predicate).ToList();
+                if(foundElems.Count != 1)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(predicate),
```
Keboo, line 105: I am not sure this is really the right exception here since it is not really the argument that is out of range, rather the result of the query that returned too many results. I would match how the LINQ Single method works, and throw an InvalidOperationException in this case.

Keboo (approved):

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
PandaMagnus: @Keboo Updated with latest commit. Also had to make some logic changes to capture an alternative use case where something could fail. Wasn't sure if you wanted to review the udpates.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
PandaMagnus: Oh, duh. Thank you for catching that. Fixed and messaging updated!

Keboo (approved):

PR #114, opened 2022-01-07
Testbuilder v2

PandaMagnus:
Buttoning up a beta version of TestBuilder v2.

Draft PR as I'm still working on unit tests AND need to type up a description for what I think this code does.

**TestBuilder.cs**
 - Remove the logic in 'ExecuteTestCase' method. Replaced with a temporary method that builds and then executes the test case (it was probably inappropriate anyway to have a 'builder' pattern that never actually builds anything.)
 - New Build() method which (that?):
   - Validates that for every input to a test block, there is a corresponding type in the DI container, and execute method override, or a test block return.
   - Validation rules around execute method overrides so we can be deterministic in how we use them.
   - A set of execute method overrides *must* be equal to or fewer in number than the corresponding execute method parameters for a give test block
   - Every override must match a type of one of the corresponding parameters
 - 
--To finish--


IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/ExampleTests.csproj
```diff
@@ -1,17 +1,30 @@
...
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
-
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <LangVersion>9.0</LangVersion>
+    <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningLevel>4</WarningLevel>
+    <NoWarn>
+      CA1303;CS1822;CS1822;<!-- Next ones are temporary. -->CS8620;CS8625;CS8600;CS8602;CS8618;CS8603;CS8604;CA1822;CA1062
```
PandaMagnus, line 10: Self note: Oops. Revisit these and remove from NoWarn if appropriate. Definitely remove comment when resolved.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/PositiveConditions/ResolverTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            TestCase tc = new TestBuilder()
+                .AddDependencyService<IExampleDataInterface, OtherExampleInterface>()
+                .AddDependencyService<IExampleDataInterface, ExampleInterface>()
+                .AddTestBlock<IncorrectInterfaceBlock>()
+                .Build();
+
+            tc.Execute();
+        }
+
+        // Need to also test a dependency that's implementing multiple interfaces
```
PandaMagnus, line 54: Self note: Double check this and remove/fix as needed.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/PositiveConditions/TestCasePropertyTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            TestCase tc = new TestBuilder()
+                .AddTestCaseId(1)
+                .Build();
+
+            // Assert
+            Assert.Equal(1, tc.TestCaseId);
+        }
+
+        // May not need below test.
+        // Still undecided if this should even be configurable.
```
PandaMagnus, line 115: YEP. Found a team not reviewing the logs at all and finally blocks were throwing errors and not cleaning up. Caused a mess in the DB. Remove comment.

PandaMagnus:
Okay, I need to just get this reviewed and merged in and tackle the rest of the cleanup with smaller, more targeted PRs. Maintaining this branch is becoming cumbersome. :-|

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/PositiveConditions/TestCasePropertyTests.cs
PandaMagnus: Will leave as-is and remove comment as part of a targeted cleanup effort.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/PositiveConditions/ResolverTests.cs
PandaMagnus: Will address this later as part of a targeted effort to shore up unit tests

IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/ExampleTests.csproj
PandaMagnus: Will leave as-is and remove comment as part of a targeted cleanup effort.

IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/ExampleTests.csproj
```diff
@@ -2,16 +2,29 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
-
+    <LangVersion>9.0</LangVersion>
```
Keboo, line 5: As per discussion would be nice to pub this up if it is a simple change. I would also bump the LangVersion to 11.

IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/ExampleTests.csproj
```diff
@@ -2,16 +2,29 @@
...
+    <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningLevel>4</WarningLevel>
```
Keboo, line 8: This looks like it is already covered in the root Directory,.Build.props file.

```suggestion
```

IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/WebDriverFactory.cs
```diff
@@ -5,7 +5,9 @@
 
 namespace ExampleTests
 {
+#pragma warning disable CA1001 // Types that own disposable fields should be disposable; Driver should be disposed by DI container, not the factory
```
Keboo, line 8: Any reason for suppressing this rather than making it disposable?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/ExampleLogger.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+
+namespace IntelliTect.TestTools.TestFramework.Tests
+{
+    public class ExampleLogger : ITestCaseLogger
```
Keboo, line 5: nit:
```suggestion
    public class ThrowingLogger : ITestCaseLogger
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBase.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Linq;
+using Xunit;
+
+namespace IntelliTect.TestTools.TestFramework.Tests
+{
+    public class TestBase
+    {
+        protected static void ValidateAggregateException(AggregateException result, int expectedInnerExceptions, params string[] messages)
+        {
+#pragma warning disable CA1062 // Validate arguments of public methods
```
Keboo, line 12: Because you can't assume that callers have nullable reference types enabled, I would add null checks rather than supress the exception.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/PositiveConditions/LoggerTests.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        public CustomLogger(TestCase tc)
+        {
+            TestCase = tc;
+        }
+        public TestCase TestCase { get; }
+
+        public string? CurrentTestBlock { get; set; }
+
+        public int Invocations { get; set; }
```
Keboo, line 161: It looks like this is only compared against `> 0`. Typically if that is all it is good for, it would  better to just make this a boolean.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestCaseTests/TestFailureTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            TestCase tc = new TestBuilder()
+                .AddDependencyService<AlwaysThrow>(new ExampleFactory().Throws)
+                .AddDependencyService<SomeDependency>()
+                .AddTestBlock<SomeTestBlock>()
+                .Build();
+
+            var ex = Assert.Throws<TestCaseException>(() => tc.Execute());
+            Assert.False(tc.Passed);
+            Assert.NotNull(ex.InnerException);
+            Assert.Equal(typeof(InvalidOperationException), ex.InnerException!.GetType());
```
Keboo, line 40: A better type check is:
```suggestion
            Assert.IsType<InvalidOperationException>(ex.InnerException);
```

IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/ExampleTests.csproj
BenjaminMichaelis, line 5:
```suggestion
    <LangVersion>11.0</LangVersion>
```
Can we bump to latest lang version?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestData/Dependencies/SimulatorClasses.cs
```diff
@@ -0,0 +1,111 @@
...
+            if (input == null) throw new ArgumentNullException(nameof(input));
+            Assert.Equal("Testing", input.Testing);
+        }
+    }
+
+    public class ExampleTestBlockWithExecuteArgForInterface : TestBlock
+    {
+        public void Execute(IExampleDataInterface input)
+        {
+            if (input == null) throw new ArgumentNullException(nameof(input));
```
Keboo, line 21:
```suggestion
            if (input is null) throw new ArgumentNullException(nameof(input));
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/IntelliTect.TestTools.TestFramework.Tests.csproj
```diff
@@ -2,15 +2,27 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <LangVersion>9.0</LangVersion>
```
BenjaminMichaelis, line 5:
```suggestion
    <LangVersion>11.0</LangVersion>
```
Same here?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestData/Dependencies/SimulatorClasses.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        private ExampleInterface Input { get; }
+    }
+
+    public class ExampleTestBlockForFactoryWithExecuteArg : TestBlock
+    {
+        public void Execute(ExampleInterface input)
+        {
+            if (input == null) throw new ArgumentNullException(nameof(input));
```
Keboo, line 55:
```suggestion
            if (input is null) throw new ArgumentNullException(nameof(input));
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestData/Dependencies/SimulatorClasses.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            Assert.Equal("Not Set", Input);
+        }
+    }
+
+    public class ExampleLoggerUsage : TestBlock
+    {
+        public void Execute(ITestCaseLogger log)
+        {
+            if (log == null) throw new ArgumentNullException(nameof(log));
```
Keboo, line 99:
```suggestion
            if (log is null) throw new ArgumentNullException(nameof(log));
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestCaseTests/TestFailureTests.cs
```diff
@@ -0,0 +1,46 @@
...
+                .AddDependencyInstance(false)
+                .AddDependencyInstance("Testing")
+                .AddTestBlock<ExampleTestBlockWithBoolReturn>()
+                .AddTestBlock<ExampleTestBlockWithMultipleDependencies>(1)
+                .Build();
+
+            var ex = Assert.Throws<TestCaseException>(() => tc.Execute());
+            Assert.False(tc.Passed);
+            Assert.NotNull(ex.InnerException);
+            Assert.Equal(typeof(DivideByZeroException), ex.InnerException!.GetType());
```
BenjaminMichaelis, line 24:
```suggestion
            Assert.Equal(typeof(DivideByZeroException), ex.InnerException.GetType());
```
I think this is unnecessary with the notNull check above on ex.InnerException

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestCaseTests/TestFailureTests.cs
BenjaminMichaelis, line 40:
```suggestion
            Assert.Equal(typeof(InvalidOperationException), ex.InnerException.GetType());
```
Same here I think

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.csproj
```diff
@@ -2,18 +2,28 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>9.0</LangVersion>
```
BenjaminMichaelis, line 5:
```suggestion
    <LangVersion>11.0</LangVersion>
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.csproj
```diff
@@ -2,18 +2,28 @@
...
     <WarningLevel>4</WarningLevel>
-    <NoWarn>1701;1702;CA1303</NoWarn>
+    <NoWarn>
+      CA1303;
+      <!-- The rest of these are temporary -->
+      <!--CS8618;
+      CS8604;
+      CS8603;
+      CS8620;
+      CS8600;-->
```
BenjaminMichaelis, line 16:
```suggestion
```
Can we remove these since there commented?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/DebugLogger.cs
```diff
@@ -0,0 +1,66 @@
...
+            string outputString = Serialize(output);
+            LogToDebug($"{TestCase.TestMethodName} - {CurrentTestBlock} - Output returns: {outputString}");
+        }
+
+        private void LogToDebug(object message)
+        {
+            System.Diagnostics.Debug.WriteLine(message);
+        }
+
+        private string Serialize(object objectToParse)
```
Keboo, line 48: I see a null check on this parameter in the catch block. Is it possible that a null gets in here?

BenjaminMichaelis (approved):

IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/WebDriverFactory.cs
PandaMagnus: Yeah, it's an odd quirk of this library. Since the factory ultimately returns an IWebDriver (which itself is disposable,) and the container that manages the dependencies should dispose of this on its own at the end of the ExecuteTestCase() method, I don't think making this disposable is the best thing to do here.

However, I think probably a refactor here will address this. I could probably get away with _not_ storing the IWebDriver at all; presumably good isolation between tests removes the need to do this. If you feel strongly this, I can do that refactor for this PR.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/ErrorConditions/ExecuteMethodErrors.cs
```diff
@@ -0,0 +1,55 @@
...
+                result.Message,
+                StringComparison.InvariantCultureIgnoreCase);
+        }
+    }
+
+    public class ExampleTestBlockWithMultipleExecuteMethods : TestBlock
+    {
+        public void Execute()
+        {
+            Assert.True(true);
```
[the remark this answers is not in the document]
PandaMagnus: I thought empty methods were warnings. I guess not! Removed in an upcoming commit.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestData/Dependencies/SimulatorClasses.cs
```diff
@@ -0,0 +1,111 @@
...
+
+namespace IntelliTect.TestTools.TestFramework.Tests.TestData.Dependencies
+{
+    // BREAK OUT INTO OWN FILES AS NEEDED
+
+    public class ExampleTestBlockWithExecuteArgForOwnType : TestBlock
+    {
+        public void Execute(ExampleInterface input)
+        {
+            if (input == null) throw new ArgumentNullException(nameof(input));
```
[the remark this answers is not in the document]
PandaMagnus: Hmmm... for some reason the following line is being flagged as potentially null. I even slapped a '!'  on input!.Testing and still get the error.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework/DebugLogger.cs
PandaMagnus: I'm not 100% sure how the nullable property propagates to consuming libraries. So... maybe? :) I think, based off of a prior comment you made, probably. If a person consumes this package without nullability enabled in their own project, they can invoke any of the public methods with their own code, so I'd be worried they could invoke with a null value. I don't want to just throw in those scenarios, but maybe a better message is needed in the catch block?

Keboo: Nullable reference work by the compiler adding attributes to your public API. Consumers with a compiler that also has the feature enabled then use these for the null analysis.

So for this, I would suggest either null checking the argument and throwing an `ArgumentNullException` if you get a null, or declare the parameter as nullable, and simply return empty string if the argument is null.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBase.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using Xunit;
+
+namespace IntelliTect.TestTools.TestFramework.Tests
+{
+    public class TestBase
+    {
+        protected static void ValidateAggregateException(AggregateException result, int expectedInnerExceptions, params string[] messages)
+        {
+            if(result is null) throw new ArgumentNullException(nameof(result));
```
Keboo, line 10: Now that your project is net6 you can use this goodness for null checks that don't include an assignment.

```suggestion
            ArgumentNullException.ThrowIfNull(result);
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/PositiveConditions/LoggerTests.cs
```diff
@@ -0,0 +1,188 @@
...
+            tc.Execute();
+        }
+    }
+
+    public class DefaultLogBlock : TestBlock
+    {
+        public void Execute()
+        {
+            Assert.NotNull(Log);
+            Assert.Equal(typeof(DebugLogger), Log!.GetType());
```
Keboo, line 105:
```suggestion
            Assert.IsType<DebugLogger>(Log)
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/PositiveConditions/LoggerTests.cs
```diff
@@ -0,0 +1,188 @@
...
+            Assert.Null(log);
+        }
+    }
+
+    public class CustomLogBlock : TestBlock
+    {
+        public void Execute()
+        {
+            Assert.NotNull(Log);
+            Assert.Equal(typeof(CustomLogger), Log!.GetType());
```
Keboo, line 143:
```suggestion
            Assert.IsType<CustomLogger>(Log)
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestCaseTests/TestFailureTests.cs
Keboo: Even better is to leverage xUnit to do these sorts of checks (combines the null and type check into one)
`Assert.IsType<DivideByZeroException>(ex.InnerException)`

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestData/Dependencies/ExampleInterface.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        string Testing { get; set; }
+    }
+
+    public class ExampleInterface : IExampleDataInterface
+    {
+        public string Testing { get; set; } = "Testing";
+    }
+
+    public class OtherExampleInterface : IExampleDataInterface
```
Keboo, line 13: nit: These both have "Interface" in their name but are classes. Is there a better name that could be used here?

IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/WebDriverFactory.cs
PandaMagnus: Refactored to remove suppression.

IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/WebDriverFactory.cs
```diff
@@ -17,14 +17,17 @@ public WebDriverFactory(string browserType)
...
-            if (_BrowserType == "Chrome")
-                _Driver = new ChromeDriver(Directory.GetCurrentDirectory());
+            IWebDriver? driver;
+            if (_BrowserType is "Chrome")
+                driver = new ChromeDriver(Directory.GetCurrentDirectory());
+            // Add other browser here.
             else
-                _Driver = new ChromeDriver(Directory.GetCurrentDirectory());
-            return _Driver;
+                driver = new ChromeDriver(Directory.GetCurrentDirectory());
```
Keboo, line 25: Do we actually need this if statement? it looks like both sides do the same thing.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBase.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using Xunit;
+
+namespace IntelliTect.TestTools.TestFramework.Tests
+{
+    public class TestBase
```
Keboo, line 6: Should this class be static since it only contains a static method, or are other classes expected to derive from it?

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestCaseTests/TestFailureTests.cs
PandaMagnus: @BenjaminMichaelis the analyzers don't read Assert.NotNull as fulfilling the null check. :( I added it to keep the red little squigglies at bay.

@Keboo change made in an upcoming push.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBase.cs
PandaMagnus: @Keboo I still get errors around the object being possibly null even after changing to the ThrowIfNull method. E.G. in this case, on the very next line the result.InnerExceptions.Count invocation gets flagged as requiring validation that 'result' is not null. Am I missing something?

IntelliTect.TestTools.TestFramework/ExampleTests/ExampleTests/WebDriverFactory.cs
PandaMagnus: I was going to add other drivers to show one way of handling cross-browser testing, buuuuut... I think I'm going to skip that. Removed.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBase.cs
PandaMagnus: Second option; a few other classes derive from this.

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/PositiveConditions/LoggerTests.cs
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    public class CustomLogBlock : TestBlock
+    {
+        public void Execute()
+        {
+            Assert.NotNull(Log);
+            Assert.IsType<CustomLogger>(Log);
+            if(Log is CustomLogger cl)
```
Keboo, line 144: I would direct cast since you know the cast will succeed.

```suggestion
            CustomLogger cl = (CutomLogger)Log;
```

IntelliTect.TestTools.TestFramework/IntelliTect.TestTools.TestFramework.Tests/TestBuilderTests/PositiveConditions/ResolverTests.cs
```diff
@@ -0,0 +1,80 @@
+﻿using IntelliTect.TestTools.TestFramework.Tests.TestData.Dependencies;
+using System;
+using Xunit;
+
+namespace IntelliTect.TestTools.TestFramework.Tests.TestBuilderTests.PositiveConditions
+{
+    public class ResolverTests
+    {
+        [Fact]
+        public void Test1()
```
Keboo, line 10: These tests could use better names

PandaMagnus: D'oh. Yeah, definitely meant to change those and then just forgot... Fixed!

Keboo (approved):

PR #120, opened 2022-01-25
Add InitializeDatabase hook to allow database seeding

COsborn2:
- Add InitializeDatabase hook taking DbContext type as parameter, evaluated immediately after database is first constructed
  - This allows for easy placement of seed data code in cases where common data may be used across multiple tests within a class or across classes (Ex: XUnit IClassFixture or ICollectionFixture)

### DIFF
--- a/IntelliTect.TestTools.Data/DatabaseFixture.cs
+++ b/IntelliTect.TestTools.Data/DatabaseFixture.cs
@@ -19,7 +19,13 @@ namespace IntelliTect.TestTools.Data
 
         private IServiceProvider ServiceProvider { get; set; }
 
-        private Dictionary<(Type, string), object> ConstructorDependencies { get; } = new Dictionary<(Type, string), object>();
+        /// <summary>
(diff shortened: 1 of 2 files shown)


IntelliTect.TestTools.Data/DatabaseFixture.cs
```diff
@@ -84,14 +90,18 @@ private TDbContext CreateNewContext()
 
             var db = (TDbContext)constructorInfo.Invoke(GetConstructorValues(constructorInfo));
 
-            if (!alreadyCreated)
+            if (alreadyCreated) return db;
+
+            await db.Database.EnsureCreatedAsync();
+
+            if (InitializeDatabase is not null)
```
Keboo, line 97: A better pattern here is to use pattern matching (pun intended). By coping it into a local variable it makes it thread-safe. As opposed to accessing the property's getter twice, which is not.
```c#
if (InitializeDatabase is { } init)
{
    await init(await CreateNewContext());
}
```

Keboo (approved):

PR #123, opened 2022-05-20
Add DatabaseFixture with support for multiple DbContexts

COsborn2:
- Add new DatabaseFixture classes with support for up to 4 DbContexts


IntelliTect.TestTools.Data/DatabaseFixture.cs
```diff
@@ -203,24 +216,34 @@ private Func<object> GetConstructorParameter<T>(ParameterInfo parameter)
             : null;
     }
 
+    internal ContextConstructionInfo GetOrAddConstructionInfo<T>()
+    {
+        var type = typeof(T);
+        if (Options.ContainsKey(type)) return Options[type];
```
Keboo, line 222: Typically you want to use `TryGetValue` rather than `ContainsKey` then access.

IntelliTect.TestTools.Data.Test/OtherSampleDbContext.cs
```diff
@@ -8,4 +8,14 @@ public class OtherSampleDbContext : DbContext
 
     public OtherSampleDbContext(DbContextOptions<OtherSampleDbContext> options) : base(options)
     { }
+}
+
+public class OtherSampleDbContextCopy : DbContext
```
Keboo, line 13: This should probably be renamed to something else.
Should it be using `IQueryable`

Keboo (approved):

PR #124, opened 2022-05-23
Fix bug causing seed data to be called twice

COsborn2:
- Fix bug causing seed data method to call itself, creating duplicate records

### DIFF
--- a/IntelliTect.TestTools.Data.Test/DatabaseFixture.SeedDatabaseOnInitialization.cs
+++ b/IntelliTect.TestTools.Data.Test/DatabaseFixture.SeedDatabaseOnInitialization.cs
@@ -38,13 +38,25 @@ public class DatabaseFixtureSeedDatabaseOnInitialization : IClassFixture<Databas
     [Fact]
     public async Task DatabaseFixtureSeed_PerformMultipleDatabaseOperations_SeedDataExistsOnce()
     {
+        await _DatabaseFixture.PerformDatabaseOperation(_ => { });
+
         await _DatabaseFixture.PerformDatabaseOperation(context =>
         {
             Assert.Equal(5, context.Persons.Count());
             Assert.False(context.Persons.Any(x => x == null));
         });
+    }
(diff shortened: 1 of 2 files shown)


IntelliTect.TestTools.Data.Test/DatabaseFixture.SeedDatabaseOnInitialization.cs
```diff
@@ -38,13 +38,25 @@ public async Task DatabaseFixtureSeed_DatabaseFixtureReused_SeedDataExists()
...
+    public async Task DatabaseFixtureSeed_DatabaseFixtureWithMultipleContexts_SeedDataExistsOnce()
+    {
+        var dbFixture = new DatabaseFixture<ReadOnlySampleDbContext, SampleDbContext>();
+        dbFixture.SetInitialize<SampleDbContext>(SeedData);
 
-        await _DatabaseFixture.PerformDatabaseOperation(context => { });
+        await dbFixture.PerformDatabaseOperation(context =>
+        {
+            Assert.Equal(5, context.Persons.Count());
+        });
```
Keboo, line 59: Should this be async?

```suggestion
        await dbFixture.PerformDatabaseOperation(async context =>
        {
            Assert.Equal(5, await context.Persons.CountAsync());
        });
```

IntelliTect.TestTools.Data/DatabaseFixture.cs
```diff
@@ -246,7 +244,8 @@ internal ContextConstructionInfo GetOrAddConstructionInfo<T>()
 
         if (_SeedComplete) return db;
 
-        _Seed?.Invoke();
+        if (seeding || _Seed is null) return db;
+        await _Seed.Invoke();
```
Keboo, line 248: Rather than passing in a flag that is only set in one place, why not simply pull this code into the `_Seed` delegate?

Keboo (approved):

PR #129, opened 2022-09-22
Selenate/upgrade selenium 4

PandaMagnus:
## Description

- Upgrade to Selenium 4
- Switch to using SearchContext as the search base to support chaining element finds

Fixes #Issue_Number (if available)

### Ensure that your pull request has followed all the steps below:
- [x] Code compilation
- [x] Created tests which fail without the change (if possible)
- [x] All tests passing
- [x] Extended the README / documentation, if necessary


Examples/IntelliTect.TestTools.Selenate.Examples/Pages/ShadowDomPage.cs
```diff
@@ -0,0 +1,22 @@
+﻿using OpenQA.Selenium;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace IntelliTect.TestTools.Selenate.Examples.Pages
```
Keboo, line 6: Given that the project was updated to C# 10, do you also want to leverage 10 features like file scoped namespaces?

IntelliTect.TestTools.Selenate.Tests/ElementHandlerTests.cs
```diff
@@ -324,7 +326,97 @@ public void GetAttributeThrowsOnTimeout()
 
             var element = SetupElementHandler(mockDriver.Object);
 
-            Assert.Throws<WebDriverTimeoutException>(() => element.GetAttribute("Test"));
+            var ex = Assert.Throws<WebDriverTimeoutException>(() => element.GetAttribute("Test"));
+            Assert.Equal(typeof(NoSuchElementException), ex.InnerException?.GetType());
+        }
+
+        [Fact]
+        public void ChainElementFindsReturnsChildElement()
```
Keboo, line 334: The name of this test is not clear to me exactly what is being tested. It seems like SetupElementHandler uses the FindElement methods?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
```diff
@@ -98,21 +111,52 @@ public bool ContainsText(string text)
...
+
+        }
+
+        public IList<IWebElement> GetElements(Func<IWebElement, bool> predicate)
         {
             IWait<IWebDriver> wait = Wait;
             wait.IgnoreExceptionTypes(typeof(NoSuchElementException));
-            return wait.Until(d =>
+
+#pragma warning disable CS8603 // Possible null reference return. Needed for Selenium retry.
```
Keboo, line 144: Is it really correct to suppress this warning? If we are going to have cases returning null, shouldn't we change the return type to reflect that?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
```diff
@@ -98,21 +111,52 @@ public bool ContainsText(string text)
...
+            return wait.Until(_ =>
             {
-                var foundElems = d.FindElements(Locator);
-                if (foundElems is null || foundElems.Count == 0) throw new NoSuchElementException($"No element found matching pattern: {Locator}");
-                var foundElem = foundElems.Where(predicate).ToList();
-                if (foundElem.Count != 1)
+                IReadOnlyCollection<IWebElement> foundElems = SearchContext.FindElements(Locator);
+                if (foundElems is null || foundElems.Count == 0) throw new NoSuchElementException($"No elements found matching pattern: {Locator}");
+                IList<IWebElement> elements = foundElems.Where(predicate).ToList();
```
Keboo, line 149: Should the predicated be applied before the if?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
```diff
@@ -98,21 +111,52 @@ public bool ContainsText(string text)
...
+
+#pragma warning disable CS8603 // Possible null reference return. Needed for Selenium retry.
+            return wait.Until(_ =>
             {
-                var foundElems = d.FindElements(Locator);
-                if (foundElems is null || foundElems.Count == 0) throw new NoSuchElementException($"No element found matching pattern: {Locator}");
-                var foundElem = foundElems.Where(predicate).ToList();
-                if (foundElem.Count != 1)
+                IReadOnlyCollection<IWebElement> foundElems = SearchContext.FindElements(Locator);
+                if (foundElems is null || foundElems.Count == 0) throw new NoSuchElementException($"No elements found matching pattern: {Locator}");
```
Keboo, line 148: The variable is declared non-nullable, so the null check should not be needed, or the variable type should be changed.
```suggestion
                if (foundElems.Count == 0) throw new NoSuchElementException($"No elements found matching pattern: {Locator}");
```

Keboo (approved):

IntelliTect.TestTools.Selenate.Tests/ElementHandlerTests.cs
PandaMagnus: Yeah, specifically this test is checking that if you do something like...

Driver.FindElement(By.Id("blah")).FindElement(By.Id("test")) then you get back that second element of ID "test". It's essentially checking that my wrapping isn't ultimately modifying the underlying Selenium behavior.

I could rename it to something like FindElementFromElementReturnsChild()?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
PandaMagnus: The challenge is Selenium here. The Until() method swallows nulls and takes that as a fail case to retry when the delegate doesn't return bool. If the delegate never returns non-null, then Selenium throws an exception. If, in this scenario, the null was ever _actually_ returned, there would be a bug in Selenium.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
PandaMagnus: I'll double check Selenium's default behavior and change as appropriate.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
PandaMagnus: Maybe a naming problem? The goal of the method is:

For a given selenium locator, for the elements Selenium returns, filter down those returned elements to some more precise list.

I'm doing the if as a final check to essentially make sure at least one element is left after doing the subsequent filtering. Does that make sense, or is something still not adding up?

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
PandaMagnus: Your suggested change is correct. :) Selenium returns an empty collection.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
PandaMagnus: Oh wait, I think I see what you're saying. I could potentially turn this into a single elements.Any() (one of your other comments confirmed Selenium's FindElements shouldn't ever return null,) check instead of two different ifs and get the same functionality. I'll think about that while I sleep.

IntelliTect.TestTools.Selenate.Tests/ElementHandlerTests.cs
Keboo: Yea I think that would have made more sense when I read it.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
Keboo: I think what I would suggest in that case is to just do `return null!;` down on line 156 and leave a comment next to it indicating that selenium treats that as failure.

Examples/IntelliTect.TestTools.Selenate.Examples/Pages/ShadowDomPage.cs
PandaMagnus: Probably. I'm doing my best to keep the example project as "best practice" as possible.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
PandaMagnus: Moved the predicate to be applied inline with the FindElements, simplified the logic to a single Any() call.

IntelliTect.TestTools.Selenate/IntelliTect.TestTools.Selenate/ElementsHandler.cs
PandaMagnus: Done and done.

Examples/IntelliTect.TestTools.Selenate.Examples/Pages/ShadowDomPage.cs
PandaMagnus: Converted to file scoped namespaces, implemented a few global usings (however I did it in a Global.cs file, not in the csproj. Not sure if there's a best practice here @Keboo ?)

Examples/IntelliTect.TestTools.Selenate.Examples/Global.cs
```diff
@@ -0,0 +1,4 @@
+﻿global using IntelliTect.TestTools.Selenate.Examples.Pages;
```
Keboo, line 1: Not suggesting a change, but wanted to make you aware of it. These global using statements can be controlled with the .csproj as well. This has the advantage of being able to leverage things like Directory.Build.props files to propagate them to multiple projects. However, I have been told by multiple people they find this un-intuitive.

Keboo (approved):

Examples/IntelliTect.TestTools.Selenate.Examples/Pages/ShadowDomPage.cs
Keboo: Left a comment on that file. There is not agreed upon best practice, but I have a minority opinion that says they should be managed in MSBuild files.

Examples/IntelliTect.TestTools.Selenate.Examples/Global.cs
PandaMagnus: I was aware, yes, but wasn't sure what the best practice was. If you have no strong objections I'll leave this as-is and look to change in the next update (which hopefully won't take nearly as long. :) )

Examples/IntelliTect.TestTools.Selenate.Examples/Pages/ShadowDomPage.cs
PandaMagnus: I'll change 'em over with the next update. Thank you!

